PR #5, opened 2021-09-08
Added process function to escape ampersand

aej11a:


with-heart (approved):
As noted by @aej11a on Discord, 3415 is an example of this problem. Validated that it renders correctly with this change. Great work!

PR #6, opened 2021-09-08
Add prettier + husky and some basic configs

xdrdak:
Adding the base prettier configuration.


with-heart (approved):
Played around with this and validated that the setup works correctly. Thanks @xdrdak!

PR #7, opened 2021-09-08
Add and Setup I18n

Canopix:
Now, there is not a Language Selector on the Website (maybe we can add one if needed). 
The website detects the language of the user and puts the correct language on the UI.

For now, only English and Spanish are available.


with-heart:
I think this is a great approach as we should want the DAO to be accessible to anyone regardless of their language. My one concern here is that items like this that change the development workflow need to be documented in one place, so that developers new to the project are able to understand what's expected of them. 

In this case, we need to make sure we set the expectation that translations are referenced via `useTranslation` and that new text blocks should be added as translations rather than static text in a component.

For Chakra UI, we use [a `CONTRIBUTING.md` file](https://github.com/chakra-ui/chakra-ui/blob/main/CONTRIBUTING.md) to document these types of things. I created #9 to discuss creating this file, and would be happy to write a section about i18n for you if you'd like!

Canopix:
Thanks for the recommendation! You're are completely right about the documenting process on "How to Add more content".  I can help you with that document if you need it. 


ps: Solving conflicts

with-heart:
> I can help you with that document if you need it.

That would be awesome! I'll get the first draft together here in a bit so we can add this to it!

with-heart (approved):

with-heart:
I'm gonna go ahead and merge this. I tend to prefer having at least 2 approvals before merging, but we can figure those kinds of standards out later!

PR #10, opened 2021-09-09
ci: Add lint task and workflow for linting on pr

xdrdak:
Another low hanging fruit that we might wanna automate and opt in sooner rather than later.


with-heart (approved):
@xdrdak I had added exactly this to my tasks list for tomorrow. You're killing it!

PR #14, opened 2021-09-10
add messaging when developer id isn't entered

cbetz:
Condition: person doesn't enter a developer id.
Before PR: person is presented with "Error."
After PR: person is presented with "Please enter a developer id"

Dev #1982 aka scrapdog reporting in!


with-heart (approved):
Nice lil improvement! If you can resolve the conflict, we'll def merge this.

cbetz:
Conflict resolved and updated to use translations. Someone may want to take a peak at the es translation; used Google Translate for it :)

PR #23, opened 2021-09-11
Added default gray to body instead of section

codingwithmanny:
**Context:**
Noticed when looking at the website at first load, it shows a white background when no id has been entered.
This is to address this to make sure the background is set on the `body` for the entire page.

**Before**
White at the bottom.
Screen Shot 2021-09-11 at 12 52 54 PM

**Proposal:**
No more white at the bottom.
Screen Shot 2021-09-11 at 12 52 39 PM

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -26,7 +26,7 @@
     -->
     <title>Developer DAO</title>
   </head>
-  <body>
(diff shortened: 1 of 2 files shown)


thomasmetta (approved):
LGTM

Issue #26, opened 2021-09-11
Empty prettier config


Dhaiwat10:
The [prettier config file](https://github.com/Developer-DAO/developer-dao/blob/main/.prettierrc.json) is...empty? 😅

Shouldn't we have some rules in place inside the config?

codingwithmanny:
Good call. Here's my initial proposal:

```json
{
  "printWidth": 80,
  "tabWidth": 2,
  "useTabs": false,
  "semi": true,
  "singleQuote": true,
  "trailingComma": "all",
  "proseWrap": "always"
}
```

Dhaiwat10:
LGTM! I'll put in a PR once we get a few more green lights :)

with-heart:
Go for it!

PR #30, opened 2021-09-11
feat: add prettier config

Dhaiwat10:
*Context*
#26 

*Changelog*
- Add the proposed prettier config
- Apply the new formatting rules to unchanged files since the pre-commit hook only applies formatting to staged/changed files


with-heart (approved):
Thank you!

PR #31, opened 2021-09-11
docs: add CONTRIBUTING.md

### LINKED ISSUE #9: Document contributor patterns/expectations
As a project grows, it's important to establish clear patterns and expectations for potential contributors in a single place. This allows new contributors to understand any special cases they need to be aware of that might affect their workflow, so they can be effective in working on the project from the get-go.

An example of a special case is #7. i18n is great and we should want that in the project, but it means that new contributors need to handle text/translations in a specific way. Having that documented could save them the trouble of not knowing that until after they open a PR.

For Chakra UI, we created [a `CONTRIBUTING.md` file](https://github.com/chakra-ui/chakra-ui/blob/main/CONTRIBUTING.md) which documents tools we use and how to use them, various npm scripts use for managing the project, the process for creating a PR, etc. 

I plan on opening a PR to create this file, but wanted to start with an issue to get feedback and see if anyone has any ideas for things to add to it.

with-heart:
Closes #9

I've been meaning to get started on this and finally made a bit of progress today. Wanted to throw up a draft PR for early feedback if any.

I'm planning to list the pieces of the current tech stack and add some basic notes about cloning the repo and opening a PR.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,4 @@
+Thanks for showing interest in contributing to the Developer DAO!
+
+This document will guide you through all the things you might need to know to
+start contributing to the project.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,3 +2,43 @@ Thanks for showing interest in contributing to the Developer DAO!
 
 This document will guide you through all the things you might need to know to
 start contributing to the project.
+
+# Development
+
+## Internationalization (i18n)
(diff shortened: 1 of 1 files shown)


with-heart:
If someone could give this a review, I think we should just go ahead and merge this. It'll evolve dynamically over time and I think there are quite a few PRs coming in that should be adding to this (#36 is an example) so it'd be better to just get it in now.

PR #36, opened 2021-09-12
Added nvmrc file to make sure correct version of node is installed

codingwithmanny:
**Context:**
Some libraries / packages when installed will give off warnings if the incorrect version of node is installed on the developer's system. In order to make sure that everyone is on the same version, I introduced an `.nvmrc` file which will help with that.

**Features:**
- New `.nvmrc` file in the root directory
- Added instructions for how to install the correct version of node under `### Local Development Environment` in the `README.md`


Dhaiwat10 (approved):
Great addition!

with-heart (approved):

PR #38, opened 2021-09-12
Add ability to direct link to your ID via id query param

CristinaSolana:
Screen Shot 2021-09-12 at 3 32 27 PM


Dhaiwat10 (approved):

with-heart (approved):
Looks great! Would love to see the value automatically synced with the URL and/or offer a "copy link" button at some point, but this is great for now!

PR #65, opened 2021-09-15
Add link to CONTRIBUTING.md

xdrdak:
Link was never fully formed. Just added the proper md syntax for linking to the right docs.


with-heart:
Good catch! @allcontributors please add @xdrdak for docs

Issue #43, opened 2021-09-13
Automatically sync URL state with current NFT id


with-heart:
Following up on #38, I think it'd be cool if the current NFT id was automatically synced with the URL. That'd make it more obvious that the functionality exists, and users could just copy the URL to share with others.

Dhaiwat10:
We should close this now

PR #71, opened 2021-09-15
feat: sync URL state with current NFT ID

Dhaiwat10:
Context — #43 

The query param — NFT ID integration was broken for some reason (try visiting https://www.developerdao.com/?id=5350). The problem was that `router.query.id` was coming up as `undefined` on the first render & hence the variable `id` was defaulting to `1`.

~~To fix this, I switched from static site generation to SSR (you cannot get query params from the server in SSG because the page is...well, static. It's only rendered once, at build time).~~ Everything now works.

### DIFF
--- a/.all-contributorsrc
+++ b/.all-contributorsrc
@@ -131,6 +131,16 @@
       "contributions": [
         "code"
       ]
+    },
+    {
+      "login": "Dhaiwat10",
+      "name": "Dhaiwat Pandya",
+      "avatar_url": "https://avatars.githubusercontent.com/u/39617427?v=4",
+      "profile": "https://github.com/Dhaiwat10",
+      "contributions": [
+        "code",
+        "review"
+      ]
     }
   ],
   "contributorsPerLine": 7,
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 
 <!-- ALL-CONTRIBUTORS-BADGE:START - Do not remove or modify this section -->
-[![All Contributors](https://img.shields.io/badge/all_contributors-12-orange.svg?style=flat-square)](#contributors-)
+[![All Contributors](https://img.shields.io/badge/all_contributors-14-orange.svg?style=flat-square)](#contributors-)
 <!-- ALL-CONTRIBUTORS-BADGE:END -->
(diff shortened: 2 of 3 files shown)


Dhaiwat10:
Would be a good idea to get this reviewed by several people since switching from completely static to SSR is kind of an important thing.

Dhaiwat10:
Also, not sure why the Vercel preview deployment is returning a 404 error. Could someone with access to the Vercel account look into this? I don't think Vercel has deployed this PR, but then again Github is showing me that all 2 checks have passed.

image

codingwithmanny:
@Dhaiwat10 can you check the other languages as well to see that it's working.

Ex: 
- `/fr/?id=1` should work
- `/es/?id=7899` should fail

Dhaiwat10:
@codingwithmanny both of them are not working. Getting the same error in both the cases. (why though?)

image

codingwithmanny:
@Dhaiwat10 looks like a case of Developer Sherlock Holmes and SO 😂 
but for real @gjsyme might be able to help, he helped build these sections.

Dhaiwat10:
image

jokes aside, we NEED tests

Dhaiwat10:
Btw https://www.developerdao.com/fr?id=5350 and https://www.developerdao.com/es?id=5350 both don't work either. :/

Dhaiwat10:
^ This makes me think that the bug was probably not introduced in this PR. Stuff was already broken (?)

Dhaiwat10:
Massive thanks to @CristinaSolana for suggesting this vanilla JS fix. No need to switch to SSR anymore. I think we can merge this now

with-heart:
Looks like the runtime error is caused by Google Translate 😬 https://github.com/facebook/react/issues/11538

with-heart (approved):
Nice work @Dhaiwat10!

Dhaiwat10:
> Looks like the runtime error is caused by Google Translate 😬 https://github.com/facebook/react/issues/11538

you're right. it's not crashing on firefox.

bloody hell.

with-heart:
> bloody hell.

Indeed! Pretty dumb lol but oh well

Issue #72, opened 2021-09-16
Some text is too thin, hard to read


Dhaiwat10:
![image](https://user-images.githubusercontent.com/39617427/133533799-30039b08-1f2d-45e8-83fa-ebde9c0117a1.png)

Anyone else feel that the underlined text is a bit too 'thin' & hard to read? I tested this on multiple devices and I feel we should definitely increase the font weight by atleast a bit.

with-heart:
Definitely agree! I left text at the default font weight in #60 so I'll attach this issue to that :)

PR #73, opened 2021-09-16
feat: replace Tailwind with Chakra UI

### LINKED ISSUE #60: Switch to Chakra UI
Issue supersedes #47 

This issue is for converting the project to use Chakra UI instead of Tailwind. See #47 for reasons why this is beneficial for us.

There's [a video and small code example available on the Chakra site](https://chakra-ui.com/docs/getting-started#nextjs) for integrating Chakra with a Next.js project.

- [ ] Wrap page renders with `ChakraProvider`
- [ ] Convert Tailwind styles to Chakra components
- [ ] Remove Tailwind-related files and dependencies

with-heart:
Closes #60, #72

This PR makes a few changes to the project:

- removes Tailwind + related dependencies + configuration
- adds Chakra UI and related dependencies
- converts the existing components and their styles from Tailwind to Chakra
	- adds minor adjustments to spacing, sizing, and color adjustments (for improved visibility and more accessible text)
- adds [`fontsource`](https://fontsource.org/) packages and imports for Inter and Source Code Pro fonts


src/components/Footer/index.js
```diff
@@ -1,25 +1,37 @@
...
 import { IconOpenSea, IconEtherscan, IconTwitter, IconDiscord } from '../Icons';
 
 function Footer() {
   const { t } = useTranslation();
 
   return (
-    <footer className="border-t border-gray-200 py-10">
-      <div className="container mx-auto max-w-6xl px-4 flex justify-center items-center flex-col">
-        <h5 className="text-sm mb-4 text-gray-600">
+    <chakra.footer borderTop="1px solid" borderColor="gray.200" py={10}>
```
with-heart, line 19: The [`chakra` factory](https://chakra-ui.com/docs/features/chakra-factory) is an object of Chakra-enabled JSX elements. It's a convenient way to render a specific element with Chakra styling attached. 

We could also use `Box as="footer"` to achieve the same effect. I think this style is slightly cleaner sometimes but it's entirely personal preference.

src/components/Footer/index.js
```diff
@@ -1,25 +1,37 @@
...
+      <VStack mx="auto" maxW="6xl" px={4} spacing={5}>
```
with-heart, line 20: The `Stack` component exists to render its `children` with an equal `spacing` between them. There are also specialized `VStack` and `HStack` components that vertically/horizontally align the components. In this case, we're vertically stacking and centering this element's `children`.

src/components/Footer/index.js
```diff
@@ -1,25 +1,37 @@
...
           {t('madeBy')}
-          <a
-            className="underline text-blue-400"
+          <chakra.a
             href="https://github.com/Developer-DAO"
+            title="Developer DAO GitHub organization"
             target="_blank"
             rel="noreferrer"
+            textDecor="underline"
+            color="blue.600"
```
with-heart, line 29: I updated this to `blue.600` to be a little more accessible. Previously it was too light which made it difficult to read.

src/components/Icons/github.js
```diff
@@ -1,30 +1,29 @@
 import React from 'react';
 import PropTypes from 'prop-types';
+import { Box } from '@chakra-ui/react';
 
-function IconGitHub({ className = '' }) {
+/**
+ * @param {import('@chakra-ui/react').BoxProps} props
+ */
```
with-heart, line 7: This is a nifty way to import and attach TypeScript types to something using `jsdoc` without having to go full-blown TypeScript. Though I think we should... 😉  

Makes it so you get autocompletion hints for Chakra props when you use `<IconGitHub>`.

src/components/Logo/index.js
```diff
@@ -1,38 +1,43 @@
...
-      width="400"
-      height="400"
-      viewBox="0 0 400 400"
-      fill="none"
-      xmlns="http://www.w3.org/2000/svg"
+    <Box
+      rounded="full"
+      objectFit="cover"
+      objectPosition="center"
+      overflow="hidden"
```
with-heart, line 14: Since we'd want the logo to always have these props attached, I moved them into the `Logo` component itself.

src/components/Logo/index.js
```diff
@@ -1,38 +1,43 @@
...
+      {...props}
     >
-      <rect width="400" height="400" fill="black" />
```
with-heart, line 14: Sizing of the icon is controlled by the `w` and `h` props to the component now.

src/layout/Page/index.js
```diff
@@ -1,16 +1,16 @@
...
 import PropTypes from 'prop-types';
-
+import { Stack } from '@chakra-ui/react';
 import Footer from '../../components/Footer';
 import Nav from '../../components/Nav';
 
 function Page({ children }) {
   return (
-    <div>
+    <Stack spacing={10}>
```
with-heart, line 9: Using `Stack` here means all of the `children` elements will be equally spaced. It also means we don't have to explicitly attach margins to each of those elements. Quite a bit less mess spread throughout.

Protip: I rely on `Stack` as often as I can. It makes for a much cleaner layout of the component, and the spacing between elements can be adjusted in a single place (the container of the elements).

src/pages/_app.js
```diff
@@ -1,7 +1,10 @@
 import React from 'react';
-import '../styles/index.css';
 import Head from 'next/head';
 import { appWithTranslation } from 'next-i18next';
+import { ChakraProvider } from '@chakra-ui/react';
+import { theme } from '../theme';
+import '@fontsource/inter/variable-full.css';
+import '@fontsource/source-code-pro/400.css';
```
with-heart, line 7: Super convenient way to bring in custom fonts without needing to deal with `@font-face`.

src/pages/_document.js
```diff
@@ -10,7 +10,7 @@ class MyDocument extends Document {
     return (
       <Html>
         <Head />
-        <body className="bg-gray-50">
```
with-heart, line 13: Background is specified in the `theme` now, so we don't need to do any specific styling here.

src/pages/index.js
```diff
@@ -26,78 +37,84 @@ function App() {
...
+            <Input
+              aria-label="Search by developer ID"
+              placeholder="Search developer id"
+              value={developerId}
+              onChange={(e) => updateDeveloperId(e.target.value)}
+              id="hero-field"
+              name="hero-field"
+              bg="white"
+            />
+          </VStack>
```
with-heart, line 54: Cleaned up a lot of the styling and nesting here. We can achieve the same result with less complexity!

src/pages/index.js
```diff
@@ -26,78 +37,84 @@ function App() {
...
-        </div>
-      </section>
+        </VStack>
+      </chakra.main>
     </PageLayout>
   );
 }
 
-function Nft(developerId) {
+function Nft(props) {
```
with-heart, line 68: This isn't really related but seeing this has been driving me nuts so now it's fixed 😁

src/pages/index.js
```diff
@@ -26,78 +37,84 @@ function App() {
...
+  if (!props.developerId) return <Text>{t('enterDeveloperId')}</Text>;
 
-  if (error || !nft) return <>{t('error')}.</>;
+  if (error || !nft) return <Text>{t('error')}.</Text>;
 
   return (
-    <>
-      <img
-        className="border-4 w-full mb-8 object-cover object-center rounded"
+    <VStack w="full" spacing={5}>
```
with-heart, line 82: It's good practice for a component to always render its own container. This allows it to be re-used in multiple contexts.

src/pages/index.js
```diff
@@ -26,78 +37,84 @@ function App() {
...
+      <chakra.img
```
with-heart, line 83: We should really be using `Image` from `next/image` but we can refactor it later.

src/theme.js
```diff
@@ -0,0 +1,17 @@
...
+      'InterVariable, -apple-system, BlinkMacSystemFont, system-ui, sans-serif',
+    body: 'InterVariable, -apple-system, BlinkMacSystemFont, system-ui, sans-serif',
+    code: '"Source Code Pro", Menlo, Monaco, Consolas, "Courier New", monospace',
+  },
+  styles: {
+    global: {
+      body: {
+        bg: 'gray.50',
+      },
+    },
```
with-heart, line 15: We can specify global styles for specific elements in here. This will be especially useful if we end up using Markdown for documentation pages or something, as we can apply a set of styles without needing to pass a bunch of props around or whatever.

src/components/Footer/index.js
codingwithmanny: Agree. I like named components vs Box as. Easy to read and find to make edits.

src/components/Icons/github.js
codingwithmanny: +1 for TS. I mean I'd make the switch myself if we're good with that direction.

src/pages/index.js
codingwithmanny: Thoughts on:

({ developerId })

codingwithmanny (approved):
LGTM. Really great work.

src/components/Icons/github.js
with-heart: I think it'd give us a lot of long-term value. Plus I love teaching so I can help get contributors up to speed if they're unfamiliar 😄

src/pages/index.js
with-heart: None!

jk lol I have no real preference other than staying consistent throughout. Some projects I destructure, some projects I stick with `props`. I'm cool with whatever we wanna do!

Issue #44, opened 2021-09-13
Add "copy link" button for linking directly to NFT id


with-heart:
Following up on #38, it'd be useful to allow users to click a button to copy the URL to the current NFT. Makes it much easier to share with others, and makes it obvious to them that the functionality exists.

Dhaiwat10:
I'll take up this one too since no one has yet 👍🏻

codingwithmanny:
Closing this because it was addressed in #79

PR #79, opened 2021-09-16
feat: add 'copy link to nft' button

Dhaiwat10:
Context: #44 

I added [Chakra UI's package for icons](https://chakra-ui.com/docs/media-and-icons/icon) too since I think we'll need to use these icons in the future as the project grows.

### DIFF
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -18,6 +18,7 @@ import {
   Button,
   Link,
   VStack,
+  useToast,
 } from '@chakra-ui/react';
 import { LinkIcon } from '@chakra-ui/icons';
 import Logo from '../components/Logo';
@@ -85,6 +86,20 @@ function App() {
 
 function Nft(props) {
   const { t } = useTranslation();
+  const toast = useToast();
+
+  const copyLinkToNFT = useCallback(() => {
+    navigator.clipboard.writeText(
+      `https://developerdao.com/?id=${props.developerId}`,
+    );
+    toast({
+      title: t('linkCopied'),
+      status: 'success',
+      duration: 5000,
+      isClosable: true,
+    });
+  }, [toast, t, props.developerId]);
+
   const { loading, error, nft } = useNft(
     DEVELOPER_DAO_CONTRACT,
     props.developerId,
@@ -129,9 +144,8 @@ function Nft(props) {
             {t('owner')}:&nbsp;{t('unclaimed')}
           </Button>
         )}
-        <Button onClick={() => copyLinkToNFT(props.developerId)}>
-          <LinkIcon />
-          &nbsp;Copy link to this NFT
+        <Button onClick={copyLinkToNFT} leftIcon={<LinkIcon />}>
+          {t('copyLinkToNFT')}
         </Button>
       </VStack>
     </VStack>
@@ -145,11 +159,6 @@ const processBase64Img = (imgStr) => {
   const processedStr = atob(base64Str).replace(' & ', ' &amp; ');
 
   return formatInfo + ',' + btoa(processedStr);
-};
-
-const copyLinkToNFT = (id) => {
-  navigator.clipboard.writeText(`https://developerdao.com/?id=${id}`);
-  alert('Link copied!');
 };
 
 export const getStaticProps = async ({ locale }) => ({
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -19,6 +19,7 @@ import {
   Link,
   VStack,
 } from '@chakra-ui/react';
+import { LinkIcon } from '@chakra-ui/icons';
 import Logo from '../components/Logo';
 import PageLayout from '../layout/Page';
 import DevName from '../components/Search/Dev/DevName';
@@ -128,6 +129,10 @@ function Nft(props) {
             {t('owner')}:&nbsp;{t('unclaimed')}
           </Button>
         )}
+        <Button onClick={() => copyLinkToNFT(props.developerId)}>
+          <LinkIcon />
+          &nbsp;Copy link to this NFT
+        </Button>
       </VStack>
     </VStack>
   );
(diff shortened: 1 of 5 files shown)


src/pages/index.js
```diff
@@ -125,6 +126,10 @@ function Nft(props) {
             {t('owner')}:&nbsp;{t('unclaimed')}
           </Button>
         )}
+        <Button onClick={() => copyLinkToNFT(props.developerId)}>
+          <LinkIcon />
+          &nbsp;Copy link to this NFT
+        </Button>
```
with-heart, line 132: `Button` has a `leftIcon` prop that makes this nice and easy :)

```suggestion
        <Button onClick={() => copyLinkToNFT(props.developerId)} leftIcon={<LinkIcon />}>
          Copy link to this NFT
        </Button>
```

src/pages/index.js
```diff
@@ -139,6 +144,11 @@ const processBase64Img = (imgStr) => {
   return formatInfo + ',' + btoa(processedStr);
 };
 
+const copyLinkToNFT = (id) => {
+  navigator.clipboard.writeText(`https://developerdao.com/?id=${id}`);
+  alert('Link copied!');
```
with-heart, line 149: Chakra has a `useToast` hook that would be great for this! https://codesandbox.io/s/busy-wiles-bpjvw

with-heart (changes requested):
Looks great! Just a few small suggestions!

src/pages/index.js
```diff
@@ -125,6 +126,10 @@ function Nft(props) {
             {t('owner')}:&nbsp;{t('unclaimed')}
           </Button>
         )}
+        <Button onClick={() => copyLinkToNFT(props.developerId)}>
+          <LinkIcon />
+          &nbsp;Copy link to this NFT
```
narbs91, line 131: Should we introduce a i18n folder with properties files to start adding static text entries to for translation?  Basically we should avoid having static text like and instead have these all stored in one central place.   Don't know what the rest of the gang thinks here?

with-heart: Oh, good catch @narbs91. We already have that, forgot about it for a minute 😅

Dhaiwat10: Good catch! I removed the static text & added the text to the translation file for English. What do we do about the rest of the languages, though? 🤔

src/pages/index.js
```diff
@@ -80,6 +81,20 @@ function App() {
...
+  const toast = useToast();
+
+  const copyLinkToNFT = useCallback(() => {
+    navigator.clipboard.writeText(
+      `https://developerdao.com/?id=${props.developerId}`,
+    );
+    toast({
+      title: t('linkCopied'),
+      status: 'success',
+      duration: 5000,
```
with-heart, line 93: imo we shouldn't pass these options and just let the toast use the defaults: 

- `status` of `info`
	- `success` is imo more for like hitting an API or things of more significant importance than copying some text to the clipboard
- `duration` of `3000` 
	- 5 seconds is a long time just to tell someone that we copied some text to the clipboard
	- 3 seconds is also a long time, but at least it's less than 5 lol

with-heart (approved):
This looks great! I left some feedback on the use of `toast` but it's just a suggestion and we can merge if we'd prefer to stick with what you've got!

src/pages/index.js
with-heart: I'm still unsure what an effective process for handling missing translations looks like. 

I don't think it's worth slowing down the dev cycle because of missing translations this early in the project's life, so for now we can just create a corresponding issue to add the missing translations (with [a task list](https://docs.github.com/en/issues/tracking-your-work-with-issues/about-task-lists) of the missing languages) and leave it to translators to catch up. 

I'll create an issue template for that so we can make it semi-automated.

src/pages/index.js
Dhaiwat10: Done!

src/pages/index.js
narbs91: Yeah, it's pretty much going to come down to getting people to translate the texts as the file grows with the project.  Those can probably be flushed out periodically as we move along via separate PR's

narbs91:
Looks awesome overall @Dhaiwat10 , great stuff!  One thing i did notice is when you cross the mobile breakpoint from desktop the text doesn't dynamically shift in size too to match the etherscan and opensea links:

Desktop:

image

Mobile:

image


Is this by design perhaps?

narbs91:
Hmm looks like the screenshots didn't serve too much use, but you can notice it if you open the dev tools and dynamically resize the window

src/pages/index.js
```diff
@@ -79,6 +81,18 @@ function App() {
 
 function Nft(props) {
   const { t } = useTranslation();
+  const toast = useToast();
+
+  const copyLinkToNFT = useCallback(() => {
+    navigator.clipboard.writeText(
+      `https://developerdao.com/?id=${props.developerId}`,
```
narbs91, line 88: Actually @Dhaiwat10 can you put this link prefix `https://developerdao.com` in the DeveloperDaoConstants file?

narbs91 (changes requested):

src/pages/index.js
Dhaiwat10: > Looks awesome overall @Dhaiwat10 , great stuff! One thing i did notice is when you cross the mobile breakpoint from desktop the text doesn't dynamically shift in size too to match the etherscan and opensea links:
> 
> Desktop:
> 
> image
> 
> Mobile:
> 
> image
> 
> Is this by design perhaps?

Yeah, it's _kind of_ by design. The text size for the owner box was reduced to allow the long address to fit on mobile screens.

Dhaiwat10: > Actually @Dhaiwat10 can you put this link prefix `https://developerdao.com` in the DeveloperDaoConstants file?

Sure thing! Thanks for suggesting this

Dhaiwat10:
Removed some unused imports, too. Let me know if everything is in order

narbs91 (approved):
Looks good to me 👍

with-heart (approved):
LGTM!

Issue #67, opened 2021-09-15
Add direct opensea link to Dev name


narbs91:
Even though we have a opensea link to the entire project itself in the footer, it would be useful to have a direct link to the searched developer after a successful search.  For example, the name "Dev 1" will link to https://opensea.io/assets/0x25ed58c027921e14d86380ea2646e3a1b5c55a8b/1

narbs91:
I'll take up this feature

aesthytik:
awesome @narbs91

codingwithmanny:
@narbs91 the current link to the dev links to an etherscan URL, what would you like to propose for showing both links? Would you have a UI change to propose?

codingwithmanny:
If `Dev#1` has a link to it, I might propose the following:

1 - If there is an owner then there show a link with an opensea icon next to it
2 - If there is no owner then there is no click through link and no opensea icon

Example for scenarios like the following:
[https://opensea.io/assets/0x25ed58c027921e14d86380ea2646e3a1b5c55a8b/7899](https://opensea.io/assets/0x25ed58c027921e14d86380ea2646e3a1b5c55a8b/7899)

<img width="2032" alt="Screen Shot 2021-09-15 at 6 22 50 AM" src="https://user-images.githubusercontent.com/318082/133416652-f8b6cdac-57ff-4e5f-aa20-8b45be2e502b.png">

narbs91:
So what I was thinking in regards to implementation @codingwithmanny was exactly what you proposed:

If the NFT has an owner, turn the the name (i.e "Dev 1") to a link to it's given opensea link (https://opensea.io/assets/0x25ed58c027921e14d86380ea2646e3a1b5c55a8b/1)

if the NFT has no owner, the name appears without the link (as it is now)

UI wise, I thinking a button from the newly added [Chakra UI library](https://chakra-ui.com/docs/form/button):

![image](https://user-images.githubusercontent.com/29411347/133546339-593e3ac0-3269-407d-a54d-f4ebf56bc72f.png)

something like that but with the opensea mini icon.  The button would open in another tab

narbs91:
Working Branch: https://github.com/narbs91/developer-dao/tree/feature/add_direct_nft_link

codingwithmanny:
Yeah we can just mark the PR as a draft and see the progress.

narbs91:
Cool, cool yeah sure.  I'll touch up the styling a bit today and open up a draft

Issue #77, opened 2021-09-16
Add direct minting


mathdroid:
It would be great if we have a button to directly mint available DEVs from the website. For example, from the currently selected Dev in the front page, and maybe elsewhere in the future. (See #76)

with-heart:
Definitely agree with this. I think it'd boost adoption if we took all of the work and effort out of minting.

Dhaiwat10:
This would be so awesome. I had to look around in far too many places to mint mine. This would break that barrier!

mathdroid:
See: https://ddao-mint.ibby.dev/

codingwithmanny:
**Suggestions:**
1 - For the first iteration only support Metamask (Unless we want to handle using [Web3Modal](https://github.com/web3modal/web3modal)
2 - abi.json file needs to be compile and stored in the repo to reference the function
3 - Error handling for unable to connect to wallet
4 - Error handling for transaction or insufficient funds
5 - Confirming a transaction can take a while, so I would suggest that once there is a transaction id returned, give a link the transaction id to eitherscan for them to check the status

nagmak:
Hey! I'd love to take on this issue if no one else is working on it already🙌🏼

thomasmetta:
abi can be found here https://etherscan.io/address/0x25ed58c027921e14d86380ea2646e3a1b5c55a8b#code

Ibby-devv:
I created https://ddao-mint.ibby.dev/ , happy to provide the code if someone wants to add it in?

nagmak:
@Ibby-devv do you want to add it in and open a PR since you've built it already? If not, I can certainly build upon what you have. I was gonna get around to this tomorrow!

PR #80, opened 2021-09-16
Feature/add direct nft link

narbs91:
Issue: #67


src/components/Search/Dev/DevName.js
```diff
@@ -0,0 +1,29 @@
...
+          rel="noreferrer"
+          title={nft.owner}
+          leftIcon={<IconOpenSea />}
+          size="lg"
+          fontSize={{ base: 'xs', sm: 'md' }}
+        >
+          {nft.name}
+        </Button>
+      ) : (
+        <h1 className="font-semibold mb-2">{nft.name}</h1>
```
fmoliveira, line 23: For a bit more concise code, I'd suggest returning this default early instead of using a ternary. I think it would read slightly easier. Just a personal opinion, though. 😃 

It would look like:

```js
function DevName({ nft, developerId }) {
  if (!nft.owner) {
    return <h1>...</h1>
  }
  
  return <Button...>...</Button>
}
```

fmoliveira (commented):
Great work on extracting the constants into a new file, I really liked that!

It all looks good to me, but I will leave the approval part for the assigned reviewers. I'm leaving a small nitpicky comment that could improve readability (in my opinion). It's optional and totally up to you to do or to not do!

Well done, @narbs91 !

src/components/Search/Dev/DevName.js
narbs91: Thanks for the review @fmoliveira !  Sounds good to me, i'll make that tweak later on today

fmoliveira:
One other thing, what do you think about making the OpenSea button have the same height as the Etherscan button, for consistency? Maybe reduce the OpenSea icon a little bit? I'm not sure if reducing it would look aesthetically good, though 🤔

codingwithmanny:
I might want to propose something here and @narbs91 you tell me what you think.

Screen Shot 2021-09-16 at 4 48 15 PM

narbs91:
> One other thing, what do you think about making the OpenSea button have the same height as the Etherscan button, for consistency? Maybe reduce the OpenSea icon a little bit? I'm not sure if reducing it would look aesthetically good, though 🤔

I'll have a play around with the styling to see how it looks but yeah it might just be a matter of shrinking the icon a tad

narbs91:
> I might want to propose something here and @narbs91 you tell me what you think.
> 
> Screen Shot 2021-09-16 at 4 48 15 PM

I think the unclaimed button is a great idea but i think it should actually be something for #77 no?  Since that button would suggest the NFT available and link off to the contract for minting when pressed.  I'm not so hot for the hover as the blues seem to be colliding a bit don't you think?

codingwithmanny:
@narbs91 I'm good with no blue hover.
For #{77} that still needs to be flushed out, so for the unclaimed, I'm also ok with just just a button (really a span) with no hover or checkmark but fits the same size as the button.
Icon should be 24px h+w for the button and padding / position of icon I think should be left aligned a bit more.

with-heart:
@fmoliveira Everyone is welcome to review a PR!

codingwithmanny (approved):
LGTM

nheingit (changes requested):
All of this looks super, thanks for this!

Only change is the `src/utils/DeveloperDaoContstants.js` where "Contstants" is misspelled.

narbs91:
> All of this looks super, thanks for this!
> 
> Only change is the `src/utils/DeveloperDaoContstants.js` where "Contstants" is misspelled.

Doh! Good catch, i'll fix that up real quick

with-heart (approved):
Looks great! Nicely done!

Issue #68, opened 2021-09-15
Add unit test framework


narbs91:
Need to add unit testing and potentially integration testing to the project

codingwithmanny:
@narbs91 would you be able to describe a bit further what you'd like to propose, just to get a better idea. React-testing-framework, snapshots, cypress e2e, playwrite, etc?

narbs91:
This is a bit of a loaded issue ;); however, i think this ticket should focus on adding unit testing for what has been implemented so far.  We can then open separate issues to add e2e tests and integration tests as well.   Framework wise, i'd have to research a bit but i figure something like [mocha](https://mochajs.org/)/[jest](https://jestjs.io/) and maybe [chai](https://www.chaijs.com/) as an assertions library.

The other thing that should be established is the "right" numbers for code/branch/line coverage.  Should we start with 80 percent coverage and tweak as we go?  i think anywhere between 70-80% coverage is usually the sweet spot but depends on the code base.

We should also, as part of this issue, add the tests as a hook and requisite to be run before a PR can be merged to main.  As well, we should add a checklist of items people should be reminded of doing as part of their PR (i.e making sure tests have been written, ensuring the app works as intended,  etc...)

codingwithmanny:
- Agree: Unit tests for now, BUT I will say not snapshots if we plan on doing e2e later. I think it would be just a lost commitment.

- Agree: 80% coverage 

- Proposal: Just Jest, maybe React Testing Library (whichever works best with NextJS)

with-heart:
> The other thing that should be established is the "right" numbers for code/branch/line coverage. Should we start with 80 percent coverage and tweak as we go? i think anywhere between 70-80% coverage is usually the sweet spot but depends on the code base.

Why should that be established?

fmoliveira:
@with-heart 
> Has anyone used Cypress at all for e2e stuff before?

I have some experience with Cypress and general best practices.

Unless anyone is willing to grab e2e tests to work on, I would definitely love to contribute with Cypress!

@codingwithmanny I find React Testing Library super handy because it forces us to test in a more user-centered way, and not only because of that, but especially because there is also Cypress Testing Library! So, with Testing Library we will have pretty much the same lingo between unit, integration and e2e tests, for example:

Jest:
```js
// get element by accessible role and label text
screen.getByRole("input", { name: "Search developer DAO id" })
```

Cypress:
```js
cy.getByRole("input", { name: "Search developer DAO id" })
```

[...]

PR #86, opened 2021-09-19
Adds unit testing framework

narbs91:
- Adds jest and react-test-library as main unit testing framework
- Adds test script and modifies build script to run tests first before actually building
- Added jest config file following next.js instructions: https://nextjs.org/docs/testing
- Added Footer unit tests
- Address issue #68


_tests_/Footer/index.test.jsx
```diff
@@ -0,0 +1,45 @@
...
+    testCommonLink(
+      openSea,
+      'https://opensea.io/collection/devs-for-revolution',
+    );
+    testCommonLink(
+      etherscan,
+      'https://etherscan.io/address/0x25ed58c027921e14d86380ea2646e3a1b5c55a8b',
+    );
+    testCommonLink(twitter, 'https://twitter.com/developer_dao');
+    testCommonLink(discord, 'https://discord.gg/zRnZpzzY');
```
with-heart, line 33: Tests like this tend to lead to brittle test suites, where changing things that aren't of real consequence to the functionality of the app causes tests to fail. 

This can lead to problems for us down the road like being unsure if our tests are providing value for us because even minor _intentional_ changes cause test failures, or discouraging refactoring because it's a lot more painful than it needs to be. 

An example of this would be us receiving a Discord vanity link (like https://discord.gg/devdao) and someone updating the `Footer` to use that new link. Not only do they need to update the link but they now need to update the test too. 

If we have to update the test every time we update the content, is the test actually providing us any value?

On top of that, there are two hidden layers of tests that already guarantee that `chakra.a` renders what it's supposed to: the Chakra test suite, and the test suite of the browser itself. 

imo what will provide us the most value from our test suite is for us to follow the `testing-library` maxim:

> The more your tests resemble the way your software is used, the more confidence they can give you.

To put it another way, what makes a test suite valuable is that it tests user behavior. An example of this would be testing that when the user changes the entered token id, the UI shows a loading state and (one loaded) displays the correct NFT.

There _are_ cases where testing something like text or an `href` makes sense, but again it's related to user behavior:
 
- When an NFT is displayed, does the owner button show the correct (and truncated) address?
- When the OpenSea link for an NFT is rendered, does the link reflect the correct dynamic `href`?

With that said, I definitely like the way you've written tests. It's very clean and is basically exactly how I'd write these tests. There's just a disconnect here between writing tests for the sake of testing things and writing tests because they're valuable tests that provide a guarantee of safety for us in building experiences for our users. Hope that makes sense!

package.json
```diff
@@ -24,13 +25,14 @@
   },
   "scripts": {
     "start": "next start",
-    "build": "next build",
+    "build": "jest && next build",
```
with-heart, line 28: I think we'd be better off leaving this as `next build`. If someone wants to build the app, they just want to build the app ya know? We can guarantee that tests pass before building in the ci pipeline at a slightly later point.

_tests_/Footer/index.test.jsx
with-heart: https://testing-library.com/docs/react-testing-library/intro/ has a bit more context to add here too

jest.config.js
```diff
@@ -0,0 +1,30 @@
+module.exports = {
```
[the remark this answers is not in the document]
narbs91: Ah cool, good call, i'll add that

_tests_/Footer/index.test.jsx
narbs91: Excellent point here and was exactly what i was looking for as i'm more akin to writing tests on the backend.   I'm definitely with you for not writing tests for the sake of writing tests (also goes back to our discussion around the test coverage percentage).  I go back and re-think the purpose of the test and follow the mantra, much appreciated @with-heart !

package.json
narbs91: Actually thinking about this one again, wouldn't it be better to have the tests run along side the build so that it pushes the developer to fix something that they potentially broke with their changes?

_tests_/Footer/index.test.jsx
with-heart: This is what I wish I had written in our discussion rather than being snarky lol glad it was a good explanation for you! Good to know that you're coming from a backend background for testing, if I think of anything else that might be helpful for someone coming from that direction of things, I'll be sure to throw it out

package.json
codingwithmanny: @narbs91 I think that would come at deployment time.
What would probably need to happen next is just have a GitHub action that runs it to show the dev that the PR is failing.

Additionally we could add it in the commit process (sometimes this can get annoying though).

narbs91: Cool, ok i'll revert this for now and we can re-visit it later on

narbs91:
Screen Shot 2021-09-19 at 11 30 36 PM

codingwithmanny:
@narbs91 clone this and noticed there were errors with useTranslation:
```sh
_tests_/Nav/index.test.jsx
  ● Console

    console.warn
      react-i18next:: You will need to pass in an i18next instance by using initReactI18next

       7 |
       8 | function Nav() {
    >  9 |   const { t } = useTranslation();
         |                 ^
      10 |
      11 |   return (
      12 |     <chakra.nav borderBottom="1px solid" borderColor="gray.200">
```

with-heart:
This is such a great start for unit tests! Let's move forward with it and adjust as necessary

PR #87, opened 2021-09-19
chore: make "missing translations" locales multiselect

with-heart:
This PR makes the `missing-locales` a multiselect so we can choose multiple locales for missing translations in the same report.


with-heart:
Just gonna merge this because it's inconsequential

Issue #92, opened 2021-09-20
Reduce NFT data response time on E2E tests


fmoliveira:
## Problem

The library `use-nft` dispatches around 12 requests to load each token's data. On my connection (80Mbps with 11ms ping), each request takes anywhere from 200ms to 700ms and the total time to load NFT data is anything between 2s to 5s. It feels slow, especially because we're not fetching this directly from the blockchain, but from traditional Web2 APIs.

Despite being slow, I've had eventual timeouts when running E2E tests. Also, I've got rate-limited quite a few times in my local environment.

## Context

The [Ethers documentation: Provider API keys](https://docs.ethers.io/v5/api-keys/) says that by default, the library uses shared community API keys for Alchemy, Etherscan and Infura APIs. These keys are only recommended for low-traffic projects and early prototyping.

It also says that if we sign up to get our own API keys for all these services, we would get a much higher request rate and faster responses as well.

While I think we would still fall on the "low-traffic" use case, for now, it sounds interesting that getting our own API keys would improve the response time. And it'd also be relieving to have our request rate increased, reducing the chance of request errors.

## Proposal

~~Continue running local environments as it is, with the shared community keys, is the best setup in my opinion – no overhead added to develop locally.~~

~~For end-to-end testing and production – would it be a good idea to sign up for our own API keys in these services and experiment if the response times would be significantly faster as the Ethers documentation states?~~

See new proposal by @codingwithmanny below: https://github.com/Developer-DAO/developer-dao/issues/92#issuecomment-923863041

codingwithmanny:
Might be worth having a script that keeps track of how many api requests are made.
The free tier for etherscan is about 100,000/day or 5/per second. Could get go over that limit every we run automated tests.

Just rethinking this it might make more sense to mock that whole thing really.

fmoliveira:
@codingwithmanny I'm actually thinking of using the API keys only on tests run on CI.

- Local environment: uses shared community keys; you're free to put your personal keys on `.env` if you want, but it's not required for the app to work
- CI environment: has its own set of keys
- Production environment: also has its own set of keys, that can be different from CI to not share usage quota

I prefer having at least one smoke test running fully e2e to ensure we're dealing correctly with the NFT APIs. I'm favourable about mocking all the other e2e tests, though!

About API usage stats – does Etherscan provide something similar on its own dashboard? I know that Infura does, but I haven't used Etherscan API myself just yet.

codingwithmanny:
[...]

PR #93, opened 2021-09-20
Adding e2e tests with Cypress

### LINKED ISSUE #75: Add e2e tests with Cypress
As discussed in another issue starting from this comment: https://github.com/Developer-DAO/developer-dao/issues/68#issuecomment-920887894 , we are willing to add e2e tests.

On that thread, we only mentioned Cypress and didn't discuss any alternatives. Apparently, that's the framework that we have more people familiar with it, and for me, personally, this is enough reason to pick Cypress – to use our current knowledge. And Cypress is very mature and has a vibrant community as well.

Anyone has something to add on this point – like proposing alternatives and comparison with Cypress? What does everyone think?

fmoliveira:
Closes #75 

- [x] Setup cypress with testing-library
- [x] Setup cypress-real-events to trigger real browser interactions instead of simulated events
- [x] Setup clipboardy plugin to access the system clipboard
- [x] Added NFT traits as alternative text to the NFT image
- [x] Changed OpenSea and Etherscan links to have more descriptive titles
- [x] Setup e2e tests to run on existing PR Workflow
- ~~[ ] Test for edge cases (invalid token shows error, unclaimed token disables owner link, etc.)~~ ([e2e tests will cover only happy paths](https://github.com/Developer-DAO/developer-dao/pull/93#discussion_r713050647))
- ~~[ ] Improve accessible query of alert container and alert close button (avoid findAll, close title could be more specific?)~~
- ~~[ ] Figure a good approach to avoid timeouts when loading NFT data (can we do it without increasing the timeout?)~~ (discussion moved to #92 )
- ~~[ ] Extract alt text helper function to a new file and add unit tests (after #86 is merged)~~

## New e2e tests

  NFT Search
    ✓ Loads first token by default (5518ms)
    ✓ Loads token id from the url (3926ms)
    ✓ Loads token with ampersand character fixed (5160ms)
    ✓ Updates url when searching for another token id (2495ms)
    ✓ Copies NFT link to clipboard (6860ms)

## Aggregate test cases to reduce execution time?

We can dramatically reduce the total execution time by reducing the number of NFT data requests only by aggregating the test cases in a different approach:

Regroup test cases as:

- Load first token by default, search for token 404 and check the URL was updated, check that ampersand character is fixed, and copy NFT link to clipboard
- Load token id from the URL

This aggregation would cut the total time by ~60%. The code of the first test case would be quite big, though, but we can relieve that by writing some custom commands or simply adding comments for each section.

I tend to prefer shorter and more focused tests, but since this would speed up our build dramatically, I'm quite willing to aggregate them. What are your thoughts about this? Should we leave the test cases small and focused, or aggregate them to cut build time as well as the local execution time for e2e tests?

Not that we have performance problems as of now, but it seems like a small compromise for a major cut in execution time.

@with-heart Thoughts on this?

## Eventual timeouts to load NFT data

I'm looking into what would be a good timeout value to prevent tests to fail due to timing out to load NFT data. The request is indeed slow so I imagine we would have to increase one of the timeouts? Any thoughts on how to better handle this issue?

Perhaps the response times could be reduced by signing up with our own Provider API keys? More about this on #92


cypress/integration/nft-search.spec.js
```diff
@@ -0,0 +1,79 @@
+/// <reference types="cypress" />
+
+import {
+  DEVELOPER_DAO_CONTRACT,
+  SITE_URL,
+} from '../../src/utils/DeveloperDaoConstants';
+
+describe('NFT Search', () => {
+  it('Loads first token by default', () => {
+    cy.visit('/');
```
fmoliveira, line 10: I'm doing `cy.visit` manually on all test cases because most of the test cases are using different URLs to set other tokens in the URL, which triggered sending requests to the default token + the one from the URL, taking more total time to run.

cypress/integration/nft-search.spec.js
```diff
@@ -0,0 +1,79 @@
...
+    cy.findByText(/Loading/).should('exist');
+    cy.findByRole('textbox', { name: /Search/ }).should('have.value', '1');
+
+    cy.findByText(/Loading/).should('not.exist');
+    cy.findByRole('img', {
+      name: 'macOS, Brackets, White Tanktop, Scala, Government, Kisumu, Divergent, JonGold',
```
fmoliveira, line 16: What are your thoughts about hard-coding the NFT traits here?

I imagine it wouldn't be an issue because it's constant on the smart contract + this helps to assert that the NFT image has accessible text.

Does it sound a bit too much? 🤔

cypress/integration/nft-search.spec.js
```diff
@@ -0,0 +1,79 @@
...
+      .realType('5555');
+    cy.findByRole('textbox', { name: /Search/ }).should('have.value', '5555');
+    cy.location('search').should('eq', '?id=5555');
+  });
+
+  it('Copies NFT link to clipboard', () => {
+    cy.visit('/?id=1234');
+    cy.findByRole('button', { name: 'Copy link to NFT' })
+      .scrollIntoView()
+      .realClick();
```
fmoliveira, line 66: Using `realClick()` here is necessary, otherwise, the copy to clipboard action leaves the document with an error `Document is not focused`. I've tried using `cy.get(something).focus()` after clicking the copy to clipboard button, but that didn't work, and it sounds like a grey area that we would need to do this as well.

Then I've first read about this library `cypress-real-events` while exploring [this code example](https://github.com/filiphric/copy-paste/blob/master/cypress/integration/spec.ts#L36-L49) by Filip Hric, and the purpose of the library sounded very sound. Instead of using Cypress's default simulated events, triggered by JavaScript, it leverages the Chrome Devtools Protocol to trigger real interactions, just like Puppeteer does. [More info on the official documentation](https://github.com/dmtrKovalenko/cypress-real-events#readme).

So I ended up replacing other usages of `click` and `type` for this library too, to better resemble the real user interactions.

src/pages/index.js
```diff
@@ -145,13 +147,32 @@ function Nft(props) {
...
+  const template = document.createElement('template');
+  template.innerHTML = svgCode;
+
+  const [svgTag] = template.content.childNodes;
+  if (!svgTag) {
+    return 'Developer NFT, failed to load text';
+  }
+
+  const textNodes = Array.from(svgTag.querySelectorAll('text'));
+  const nftTraits = textNodes.map((node) => node.textContent).join(', ');
```
fmoliveira, line 173: I've added alt text to the NFT image so I could test it for accessibility. Kinda out of scope for my task, but I thought it was ok since it's for a good cause.

My first attempt at this alt text generation was to simplify sanitize the SVG code from all HTML input using the library [sanitize-html](https://bundlephobia.com/package/sanitize-html@2.5.1), which worked flawlessly: the tags were stripped and the text content left there as I wished. But the library weighs 45KB minified+gzipped, and that's too much.

I've tried sanitizing with [dompurify](https://bundlephobia.com/package/dompurify@2.3.2) which is significantly smaller (7KB), but this library behaves differently. It strips the text content together with the sanitized tags.

Then this approach only relies in the modern HTML tag `template` which does not render the underlying HTML code until we clone it into another tag, and allow us to do regular element queries.

cypress/integration/nft-search.spec.js
fmoliveira: _ sounded very sound _ was not intended lol

cypress/integration/nft-search.spec.js
codingwithmanny: We could technically. I'm trying to figure out how big that file would be.

grothem:
Nice! What do you think about adding cypress specific attributes that we can target just for these tests, instead of using `findByRole` and `findByText`? We could add `data-cy=search` or something similar to the id search input and then use that in the selector in the cypress test. That way the tests could be more stable as we add more features and more tests.

cypress/integration/nft-search.spec.js
codingwithmanny: @fmoliveira
Oh yeah, actually that makes more sense. Forgot that we don't need all permutations 🤦

fmoliveira:
> Nice! What do you think about adding cypress specific attributes that we can target just for these tests, instead of using `findByRole` and `findByText`? We could add `data-cy=search` or something similar to the id search input and then use that in the selector in the cypress test. That way the tests could be more stable as we add more features and more tests.

Hey @grothem , thanks for the comment!

I've used the testing-library queries because that's the approach we discussed on #68 and #75. Both approaches have their trade-offs, but from the discussion we had, the majority seemed to be more on board with the Testing Library maxim:

> The more your tests resemble the way your software is used, the more confidence they can give you.

(source: [Guiding Principles](https://testing-library.com/docs/guiding-principles/))

And according to the testing-library's guiding principles, using test IDs like `data-cy` violates that maxim because the user does not look for test IDs when using the UI (so we may unintentionally create false positives when using test IDs):

> In the spirit of the guiding principles, it is recommended to use this only after the other queries don't work for your use case. Using data-testid attributes do not resemble how your software is used and should be avoided if possible.

 (source: [ByTestId docs](https://testing-library.com/docs/queries/bytestid/#api))

I certainly understand your concern, anyway! Any change in the label of the search input would need the test to be adjusted, which is an indication of a brittle test. That's why I'm not querying for the complete label text, but rather only for the meaningful keyword that I believe a user would look for – the word "search". So, I believe this would check two boxes:

- [x] it resembles how the software is used (user looks for an input that has "search" in its label)
- [x] probably won't break if the label text is changed, because we shouldn't remove the "search" word from the label

(note: I should probably make that query case-insensitive, though, and try to make queries for the other links more open instead of looking for the whole string as well)

What are your thoughts about this all? Please note that I'm not being defensive of my design choice at all! I'm referring back to why we choose to use Testing Library and expanded with my thoughts about both approaches. I'm more than happy to iterate on a different approach and strive to make these tests as stable as possible. Please feel free to add any other thoughts or examples that you may have.

@with-heart @codingwithmanny @narbs91 Any thoughts?

@grothem Thank you again for commenting!

fmoliveira:
@grothem @with-heart @codingwithmanny @narbs91 

PS: I'm not at all trying to make a big deal of this with my long reply, but since this is a foundation for our E2E tests, I think it's important that we're happy with these choices and document them for our future selves. Happy to continue iterating on this as everyone sees fit!

grothem:
@fmoliveira no problem, thanks for linking to the discussion and testing-library. I didn't know that there were previous discussion on that, sorry! I've used Cypress in the past and they suggest using the `data-cy` attributes, which is why I made the comment. But makes sense what you've got here!

fmoliveira:
@grothem Sounds good! I understand that Testing Library adds a different nuance to what Cypress suggest for queries, and I think the best part of this nuance is that our queries will be quite similar among unit, integration and e2e tests. So it will also help to onboard new folks due to this similarity.

As a side note, I do think I have to review the texts in some of the queries, trying to query only for the main keywords and make them case-insensitive. I will push these patches tonight.

narbs91:
> > Nice! What do you think about adding cypress specific attributes that we can target just for these tests, instead of using `findByRole` and `findByText`? We could add `data-cy=search` or something similar to the id search input and then use that in the selector in the cypress test. That way the tests could be more stable as we add more features and more tests.
> 
> Hey @grothem , thanks for the comment!
> 
> I've used the testing-library queries because that's the approach we discussed on #68 and #75. Both approaches have their trade-offs, but from the discussion we had, the majority seemed to be more on board with the Testing Library maxim:
> 
> > The more your tests resemble the way your software is used, the more confidence they can give you.
> 
> (source: [Guiding Principles](https://testing-library.com/docs/guiding-principles/))
> 
> And according to the testing-library's guiding principles, using test IDs like `data-cy` violates that maxim because the user does not look for test IDs when using the UI (so we may unintentionally create false positives when using test IDs):
> 
> > In the spirit of the guiding principles, it is recommended to use this only after the other queries don't work for your use case. Using data-testid attributes do not resemble how your software is used and should be avoided if possible.
> 
> (source: [ByTestId docs](https://testing-library.com/docs/queries/bytestid/#api))
> 
> I certainly understand your concern, anyway! Any change in the label of the search input would need the test to be adjusted, which is an indication of a brittle test. That's why I'm not querying for the complete label text, but rather only for the meaningful keyword that I believe a user would look for – the word "search". So, I believe this would check two boxes:
> 
> * [x]  it resembles how the software is used (user looks for an input that has "search" in its label)
> * [x]  probably won't break if the label text is changed, because we shouldn't remove the "search" word from the label
> 
> (note: I should probably make that query case-insensitive, though, and try to make queries for the other links more open instead of looking for the whole string as well)
> 
> What are your thoughts about this all? Please note that I'm not being defensive of my design choice at all! I'm referring back to why we choose to use Testing Library and expanded with my thoughts about both approaches. I'm more than happy to iterate on a different approach and strive to make these tests as stable as possible. Please feel free to add any other thoughts or examples that you may have.
> 
> @with-heart @codingwithmanny @narbs91 Any thoughts?
> 
> @grothem Thank you again for commenting!

I'm on the side of sticking to the testing-libraries way of doing things with the tests to keep it consistent throughout the project.  The one thing that I noticed while working on the unit tests is that I think there's some room for improvement around the way we've set up our components to be more "test friendly".  What i mean by that is that most of our react elements aren't id'd or classed in a way that makes finding said elements easy in the tests.  In my past experience, what's worked the best is assigning unique id's (or classes) to elements so that we can more easily locate them.  For example, we have a couple of buttons whose `title` attributes are the same and just so happen to be on the same page together.  So when you do `getAllByText` you're going to get both buttons but unless you know the layout of the page specifically, you kinda have to guess which button is which.  I mean it's not a huge deal now but something we should look to addressing as we continue to work on new features and refactor what we have so far.

cypress/integration/nft-search.spec.js
```diff
@@ -0,0 +1,79 @@
...
+    cy.findByRole('textbox', { name: /Search/ }).should('have.value', '1');
+
+    cy.findByText(/Loading/).should('not.exist');
+    cy.findByRole('img', {
+      name: 'macOS, Brackets, White Tanktop, Scala, Government, Kisumu, Divergent, JonGold',
+    });
+    cy.findByRole('link', { name: 'View NFT on OpenSea' }).should(
+      'have.attr',
+      'href',
+      `https://opensea.io/assets/${DEVELOPER_DAO_CONTRACT}/1`,
```
narbs91, line 21: Might be better to define the link prefix as a global const at the top of the class so you can reference it from one spot

cypress/integration/nft-search.spec.js
```diff
@@ -0,0 +1,79 @@
...
+    );
+    cy.findByRole('link', { name: 'View owner on Etherscan' })
+      .should('have.attr', 'href')
+      .and('match', /^https:\/\/etherscan.io\/address\/0x\w+$/);
```
narbs91, line 25: Is there a reason you regex match here and not for the opensea link above?

src/components/Search/Dev/DevName.js
```diff
@@ -11,7 +11,7 @@ function DevName({ nft, developerId }) {
         href={`${OPENSEA_DIRECT_LINK_PREFIX}/${developerId}`}
         target="_blank"
         rel="noreferrer"
-        title={nft.owner}
+        title="View NFT on OpenSea"
```
narbs91, line 14: So for something like this, I would recommend you add an id to this element and call it something like `view-nft-open-sea` and then you can more easily query for it.  Thought?

src/pages/index.js
```diff
@@ -125,7 +127,7 @@ function Nft(props) {
             href={`${ETHER_SCAN_LINK_PREFIX}/${nft.owner}`}
             target="_blank"
             rel="noreferrer"
-            title={nft.owner || t('unclaimed')}
+            title="View owner on Etherscan"
```
narbs91, line 130: Same deal here as the comment prior

src/pages/index.js
narbs91: Just curious, what would the alt-text read for the NFT?

cypress/integration/nft-search.spec.js
```diff
@@ -0,0 +1,79 @@
+/// <reference types="cypress" />
+
+import {
+  DEVELOPER_DAO_CONTRACT,
+  SITE_URL,
+} from '../../src/utils/DeveloperDaoConstants';
+
+describe('NFT Search', () => {
```
narbs91, line 8: Are we just focusing on happy path this go around or did we want to also test when things don't load properly?  Like when we get an error from the useNft function that actually fetches the NFT or if someone fudges the id query parameter and adds `id=avd` for example:

image

narbs91 (changes requested):

fmoliveira:
@narbs91 Thanks for the comments!

I feel you about improving the a11y titles to simplify querying. Thinking from an accessibility viewpoint, we should definitely modify these titles because they will be confusing to screenreader users. For example, a screenreader allows the user to hear the list of all links on the page. If the titles between multiple links are equal, a screen reader user may think they lead to the same place and click whatever of them, and land into an unexpected link – for example, landing on the OpenSea collection page instead of the page of a particular token. So I think this is the core issue that we have to address – if are using a screen reader and not looking at the UI, how do we feel about the link titles ourselves? Can we navigate properly?

Also, querying elements by id or class names is not recommended both by [cypress](https://docs.cypress.io/guides/references/best-practices) and [testing-library](https://testing-library.com/docs/queries/bytestid/#api), because they are implementation details and too coupled to what's happening under the hood, things that the user doesn't interact directly with. So this falls back into writing more descriptive a11y titles. That's why I love using findByRole, it really helps thinking about a11y while writing tests as well.

cypress/integration/nft-search.spec.js
fmoliveira: Because I'm not considering the account ID relevant here, I used a regex match to ensure it's at least properly formatted. But maybe it's too much of a stretch, I could simply use a `contains` instead so it's easier to reuse the link from constants. 🤔

cypress/integration/nft-search.spec.js
fmoliveira: Good catch, this only covered the happy path. I'm going to address edge cases as well, thanks for the note!

narbs91:
> @narbs91 Thanks for the comments!
> 
> I feel you about improving the a11y titles to simplify querying. Thinking from an accessibility viewpoint, we should definitely modify these titles because they will be confusing to screenreader users. For example, a screenreader allows the user to hear the list of all links on the page. If the titles between multiple links are equal, a screen reader user may think they lead to the same place and click whatever of them, and land into an unexpected link – for example, landing on the OpenSea collection page instead of the page of a particular token. So I think this is the core issue that we have to address – if are using a screen reader and not looking at the UI, how do we feel about the link titles ourselves? Can we navigate properly?
> 
> Also, querying elements by id or class names is not recommended both by [cypress](https://docs.cypress.io/guides/references/best-practices) and [testing-library](https://testing-library.com/docs/queries/bytestid/#api), because they are implementation details and too coupled to what's happening under the hood, things that the user doesn't interact directly with. So this falls back into writing more descriptive a11y titles. That's why I love using findByRole, it really helps thinking about a11y while writing tests as well.

Ah sorry, yes totally agree on the title changes (do what i mean, not what i say ;) ).  I didn't know that searching by id/class was discouraged as from a testing point of view, at least personally, it makes finding things easier; however, if that's the way to do things i'm all good with the way we're finding things now, I'm not afraid to embrace new(better) ways of doing things

src/components/Search/Dev/DevName.js
fmoliveira: Using ids for testing is not a recommended practice both on [cypress](https://docs.cypress.io/guides/references/best-practices) and [testing library](https://testing-library.com/docs/queries/bytestid/#api), because it's an implementation detail.

On the other hand, adding a proper link description ensures we're making this link more accessible for screen reader users. Imagine using a screen reader to hear the title of the links, and one of the links say a huge hexadecimal number, spelt character-by-character – can the user figure out what does that mean? And is that useful in the context of this link? Since the goal of the link is to view the NFT on OpenSea, I changed the link title to match this intent. Simpler for the test to find this element, and useful for screen reader users.

cypress/integration/nft-search.spec.js
narbs91: Ah I see, I was just more curious on why the two approaches but that makes sense.  I would think the `contains` would probs be a smidge easier to read and easier to reuse the constants like you said.

fmoliveira:
@narbs91 For some reason I couldn't reply one of your comments so I'm quoting it here

> Just curious, what would the alt-text read for the NFT?

It reads all the NFT traits, separated by commas. I think I will just extract this out to a helper file and add a unit test to make it better documented.

With this alt text, we can ensure that screen reader users will be able to have the same experience as all other users do, like:

```
cy.findByRole('img', {
      name: 'macOS, Brackets, White Tanktop, Scala, Government, Kisumu, Divergent, JonGold',
    });
```

narbs91:
> @narbs91 For some reason I couldn't reply one of your comments so I'm quoting it here
> 
> > Just curious, what would the alt-text read for the NFT?
> 
> It reads all the NFT traits, separated by commas. I think I will just extract this out to a helper file and add a unit test to make it better documented.
> 
> With this alt text, we can ensure that screen reader users will be able to have the same experience as all other users do, like:
> 
> ```
> cy.findByRole('img', {
>       name: 'macOS, Brackets, White Tanktop, Scala, Government, Kisumu, Divergent, JonGold',
>     });
> ```

Maybe we can prepend `Developer traits: ` to the alt text so that it gives the user more context as the screen reader is reading it out?

fmoliveira:
@narbs91 100% agree, great suggestion!

fmoliveira:
@narbs91 And thank you so much for the comments! I gotta take a look at the unit tests you're writing too, excited to see the tests come to life! 😃

narbs91:
> @narbs91 And thank you so much for the comments! I gotta take a look at the unit tests you're writing too, excited to see the tests come to life! 😃

No prob at all :).  Yes, please have a look as i've just pushed some changes to mock out the translations library causing warns and let me know if you see a better way to do things then i've done (still kinda new to frontend testing with react).

src/pages/index.js
fmoliveira: It reads all the NFT traits, separated by commas. I think I will just extract this out to a helper file and add a unit test to make it better documented.

With this alt text, we can ensure that screen reader users will be able to have the same experience as all other users do, like:

```
cy.findByRole('img', {
      name: 'macOS, Brackets, White Tanktop, Scala, Government, Kisumu, Divergent, JonGold',
    });
```

(figured why I couldn't reply and [copying my answer](https://github.com/Developer-DAO/developer-dao/pull/93#issuecomment-923506924) here for posterity)

cypress/support/commands.js
```diff
@@ -0,0 +1,30 @@
...
+    .should('have.attr', 'href')
+    .should('eq', `${OPENSEA_DIRECT_LINK_PREFIX}/${token}`);
+});
+
+Cypress.Commands.add('findEtherscanLink', () => {
+  cy.findByRole('link', { name: /view .+ etherscan/i })
+    .should('exist')
+    .should('have.attr', 'href')
+    .should('contain', ETHER_SCAN_LINK_PREFIX);
+});
```
with-heart, line 30: These make me wonder if we need to define our e2e, integration, and unit test layers a little more. imo e2e (and integration, to a lesser extent) tests are for testing happy path behaviors (like, does searching for an NFT id display the actual NFT?), while things like checking `input` values or link `href`s and making sure the NFT component renders with ampersands are better served by individual component unit tests.

Not sure it matters at this stage! Just wanted to share that thought.

src/components/Search/Dev/DevName.js
with-heart: Should we use a translation here?

cypress/support/commands.js
fmoliveira: I hadn't had that thought, but it makes sense!

narbs91 (approved):
Looks good to me @fmoliveira , awesome job!

fmoliveira:
@narbs91 Thank you!! 🙌🏻 

@with-heart I've cherry-picked your change from #102 here to avoid adding new i18n local mocks.

with-heart:
Done! Gonna rebase and merge!

PR #103, opened 2021-09-21
POC: refactor: flatten project file structure

with-heart:
### What does it do?

This is a proof-of-concept PR demonstrating a flattened file structure for the project.

For how simple our project is at this point, we have imo an unnecessarily complex file structure. This POC demonstrates what a flattened file structure could look like. 

### Changes

Tree structure is viewable here: https://github.com/Developer-DAO/developer-dao/tree/flat-file-structure/src

- co-located test files
	- easier to find the test file for a thing
	- imports are much more simple (`./DevName` vs `../../../src/components/Search/Dev/DevName`)
- moved `_tests_/mocks/DeveloperNFT` to `src/test-data.js`
- moved `_tests_/utils/testCommons.js` to `src/test-utils.js`

Looking for thoughts+feedback!

cc @narbs91


narbs91:
I guess taking a current look at things, I agree that our project structure would benefit from a flat structure and  simplify things; however, would it not be better to set a precedent with how our project structure should look as we continue to build?  I guess the word 'should' here might be a tad bit strong at this point in the projects evolution lol

with-heart:
> I guess taking a current look at things, I agree that our project structure would benefit from a flat structure and simplify things; however, would it not be better to set a precedent with how our project structure should look as we continue to build? I guess the word 'should' here might be a tad bit strong at this point in the projects evolution lol

Ye! Been meaning to write more about how I think about maintaining an open source project to explain this kinda stuff, but that last sentence is definitely key.

I'm not anti-process or anti-file-structure-changes by any means, I just think it's important for us to make sure new processes or changes that make things more complex have a good reason behind them and are documented. 

The more complex we make things, the more there is for new contributors to have to understand in order to start working, so I think it's healthy to at least make sure we justify the complexities we want.

There's also kind of a like "monopoly of the present" effect that can happen too, where decisions we make today tend to affect people long after we've moved on from the project. 

Especially with how young the project is and how few contributors we have compared to how many I think we'll have once we mature, it's kind of unfair for us to lay down too many laws and processes that others that come later didn't get a say in but have to deal with.

But we can change anything, so it's not like it's a _massive problem_ that we can never get out of, just some things I like to keep in mind :)

narbs91:
Totally agree with what you said @with-heart, we can flatten out the structure to reduce complexity to allow an easier contribution entry.  I really want to avoid that "monopoly of the present" effect and let the community as a whole drive the decisions cause that's the whole spirit of this grand marathon after all (at least to me anyhow).

codingwithmanny:
@with-heart @narbs91 
This is a personal preference, but I sort of like the tests within the scope of their functionality that way everything that pertains to that element, or piece of functionality is nearby in terms of dealing with it as a organization point of it.

Not oppose to the other format, but as personally I like it as it is.

with-heart:
Whoops I missed this!

> This is a personal preference, but I sort of like the tests within the scope of their functionality that way everything that pertains to that element, or piece of functionality is nearby in terms of dealing with it as a organization point of it.

I'm not following what you mean here!

with-heart:
Doesn't matter! This kinda stuff isn't relevant rn!

PR #105, opened 2021-09-21
fix: Validate token number from url parameter

### LINKED ISSUE #98: Validate token number from url parameter
The search input already has a validation that disallows typing token numbers greater than 8000, our maximum token number.

However, we can still type an arbitrary number in the URL and the page shows as if the token existed and is available to be claimed. This may look like trolling for people familiar with the token limit, but otherwise, having an error message saying something like "Invalid token number" would be helpful.

Steps to reproduce:
- Access https://www.developerdao.com/?id=150258
- The NFT image is rendered, but it should error instead

Screen Shot 2021-09-20 at 11 21 44 PM

moiz-lakkadkutta:
### What does it do?
If someone passes an id > 8000 in query params, the input will not be filled with that id & instead it will default to 1

I tried to add an alert but ran into some errors so i have commented it.

--

### Relevant screenshots/gifs
--

### Does it close any issues?

Closes #98


src/pages/index.js
```diff
@@ -46,7 +46,13 @@ function App() {
   function getSearchID() {
     if (process.browser) {
       const search = window.location.search;
-      return new URLSearchParams(search).get('id') || 1;
+      const val = new URLSearchParams(search).get('id') || 1;
+      if (val > 8000) {
```
narbs91, line 50: You'll probably want to shift this to also use a regex check of sorts since technically someone could pass in something like `abc` as a query param; however, i don't know if we actually care as the error message will show up if they do:

image

You can perhaps look to add a validity checker method in the class to enclose the details of the check, i.e

```
function isQueryParamValid(val) {
 .....
}
```

Then you can have you code revised just a tad to be:

```
if (process.browser) {
      const search = window.location.search;
      const val = new URLSearchParams(search).get('id') || 1;
      if (isQueryParamValid(val)) {
        return val;
      }
}
return 1;  // You'll always return one on a non-valid id in this case
```

src/pages/index.js
```diff
@@ -46,7 +46,13 @@ function App() {
...
+        // alert(t('invalidToken'))
```
narbs91, line 51: Let's not leave commented code in master

narbs91 (changes requested):

public/locales/en/common.json
```diff
@@ -10,5 +10,6 @@
   "enterDeveloperId": "Please enter a Developer DAO id",
   "hosting": "Hosting by:",
   "copyLinkToNFT": "Copy link to NFT",
-  "linkCopied": "Link copied to clipboard"
+  "linkCopied": "Link copied to clipboard",
+  "invalidToken": "Invalid token ID"
```
narbs91, line 14: Perhaps you can use this as a error message when the id isn't valid (instead of the alert)

fmoliveira (changes requested):
The id in the URL is different from the id in the search input, can we update it in the router as well?

So, if the user types an invalid id in the URL and we display token # 1 instead, the id in the URL should also reflect token # 1.

image

src/pages/index.js
moiz-lakkadkutta: yeah sure

public/locales/en/common.json
moiz-lakkadkutta: Great

src/pages/index.js
```diff
@@ -102,7 +113,7 @@ function Nft(props) {
 
   if (!props.developerId) return <Text>{t('enterDeveloperId')}</Text>;
 
-  if (error || !nft) return <Text>{t('error')}.</Text>;
+  if (error || !nft) return <Text>{t('invalidToken')}.</Text>;
```
with-heart, line 116: We have a problem in the logic of the system here:

- `getSearchId` prevents invalid value from query params
	- `developerId` is initially set using `getSearchId`
	- `getSearchId` sets the id to `1` if it's invalid
- `updateDeveloperId` prevents invalid value from change event
	- `updateDeveloperId` is connected to the `Input` through `onChange={updateDeveloperId}`
	- `updateDeveloperId` only updates `developerId` if the new value is `<= 8000`

Since we prevent the query params and user input from setting an invalid id, we'll never see this error due to an invalid token.

Not sure what to do about that situation. We could allow an invalid `developerId` so we can show an error message that allows the user to understand the limitation. Or we could just drop the `invalidToken` message and continue to prevent all invalid values.

Thoughts?

cc @codingwithmanny

with-heart (changes requested):

src/pages/index.js
with-heart: Personally, I don't think _preventing_ invalid user input is helpful. It's more informative to allow it and then display an error state so the user can understand _why_ the value is invalid.

codingwithmanny: @with-heart Please correct me If I'm misinterpreting this.

We would need a not found, error, or invalid developer id state, that way we don't have to account for all these conditions, just `<div>{showNftResult(developerId)}</div>` and in it would just handle if it's a valid number to show the correct result or in an invalid one to show an error message.

This would also solve the issue of querying a token with etherjs and the request fails because of API limits.

moiz-lakkadkutta: What if we just change it to -1 if its invalid and then show "Token is Invalid".

moiz-lakkadkutta: dev_dao

narbs91: I would be on the side of showing the id that caused the issue rather then displaying -1, what do you think @with-heart / @codingwithmanny ?

moiz-lakkadkutta: we could do that but the problem with that will be if id is changed from url and given a value like "abc" that will give error of "Invalid token" but input parameter will also change to "abc" and it wont change from input because the next value "ab" is not a number.


image

moiz-lakkadkutta: Other than that only solution i can think of is adding a submit button and remove restriction on input box so it can show proper error if something goes wrong.

codingwithmanny: @with-heart @moiz-lakkadkutta @narbs91 
This is might not be a popular solution, might be a tag more work, but we could add routing, and the params only accept numbers, and if the value is invalid (-1, 8001, abc), then we just show a 404 page.

This would also make some of the meta data for that id unique when sharing it too.
ex: https://developerdao.com/id/1234 and then when sharing it on social media, we could maybe even show something like the SVG graphic.

moiz-lakkadkutta: @codingwithmanny 
for that to a button or something like has to be added.

codingwithmanny: Yeah we'd probably need to adjust the button as well.

codingwithmanny: @moiz-lakkadkutta for the button, I think we already have one, or are you talking about a 404 page?
PS Feel free to hit me up on discord (same username)

moiz-lakkadkutta: @codingwithmanny  when i saw the deployment of my PR (Below by vercel bot)it goes to 404 page for every id but when i ran it on my local host it ran just fine. I don't understand why is that, can you help me out a bit?

codingwithmanny:
@moiz-lakkadkutta 

Just looked the deployment, and I noticed a few bugs, but before I list them I'm getting worried we might be working on something that might get changed in #114 , but who knows when that will be implemented.

Here's the list of recommendations:

1 - Re: Search Button: I think we still want the live search on the main home page because it's just much easier to look up something without having to click a submit button, maybe I forgot to communicate that, so my bad.

2 - Re: /id/# - I think it's the right approach but I keep getting 404s, it also looks like it's not working correctly for the other languages as well, when you try `/es/id/1`

moiz-lakkadkutta:
@codingwithmanny 

but with that there will still be a bug say if user changes url to "/id/#abc" then the input parameter will also change and user won't be able to change it from input as the resulting string won't be less than 8000 or a number ("ab") for that matter.

codingwithmanny:
@moiz-lakkadkutta 

Sorry I meant that we wouldn't up the the #param based on the user's input in the input field. So it would be:

**Scenario 1:**
- User goes to website
- User enters a value of 6269 in the text field
- User sees immediate result (nothing is updated in the top navigation url)
- User can still copy the link, which will be `https://developerdao.com/id/6269`

**Scenario 2:**
- User goes to direct link `https://developerdao.com/es/id/6269`
- User sees the input pre-filled with `6269`
- User sees the graphic for `6269`
- User enters a new number `1234` (nothing is updated in the top navigation url)
- User sees the graphic for `1234`
- User clicks copy link, which will be `https://developerdao.com/id/1234`

**Scenario 3:**
- User go to either `https://developerdao.com/es/id/8001`, `https://developerdao.com/id/-1`, `https://developerdao.com/fr/id/abcd`
- User is shown a 404 page

moiz-lakkadkutta:
@codingwithmanny  
please ignore this PR.

moiz-lakkadkutta:
@codingwithmanny 

i wanted to ask that on my localhost it runs fine but the deployment by vercel bot shows 404 page for every id ("/id/#").

image

image

with-heart:
Sorry that I haven't checked in on this recently. I've been on vacation this week so my time online has been fairly minimal. Not my intent to leave this hanging, and I'll review it and the latest discussion tomorrow.

moiz-lakkadkutta:
@narbs91 
can you have a look at this

I am getting this warning

image

and a possible solution can be 

https://react.i18next.com/latest/usetranslation-hook#not-using-suspense

image

narbs91:
> @narbs91
> can you have a look at this
> 
> I am getting this warning
> 
> image
> 
> and a possible solution can be
> 
> https://react.i18next.com/latest/usetranslation-hook#not-using-suspense
> 
> image

Heya sorry for the late reply, so to my knowledge that warning is because the `serverSideTranslations` needs to be included/setup on that page so it can actually do the translations according to the [next-i18next documentation](https://github.com/isaachinman/next-i18next#serversidetranslations).  I wonder if it's because that 404 page is something that's "under the hood" of nextjs that it doesn't include it when it renders.  I'm fairly new to next.js so perhaps @with-heart and/or @codingwithmanny can speak better to what's going on here

narbs91:
@moiz-lakkadkutta I think you're issue with the dynamic routing may have to do with how you've defined the `getStaticPaths` and how next.js treats the routing in development mode and in production.  According to there [documentation](https://nextjs.org/learn/basics/dynamic-routes/dynamic-routes-details), 

```
Development v.s. Production
In development (npm run dev or yarn dev), getStaticPaths runs on every request.
In production, getStaticPaths runs at build time.
```

So locally, every request you run successfully displays the page; however, with the deployed version that method only runs once when the app is built and no more.  Another thing which may aid in the issues here is perhaps we need to set the `id` in the `getStaticProps` since you've set the `fallback` to blocking.  This is all spitballing at this point as i'm literally learning this all in the moment haha

moiz-lakkadkutta:
> @moiz-lakkadkutta I think you're issue with the dynamic routing may have to do with how you've defined the `getStaticPaths` and how next.js treats the routing in development mode and in production. According to their [documentation](https://nextjs.org/learn/basics/dynamic-routes/dynamic-routes-details),
> 
> ```
> Development v.s. Production
> In development (npm run dev or yarn dev), getStaticPaths runs on every request.
> In production, getStaticPaths runs at build time.
> ```
> 
> So locally, every request you run successfully displays the page; however, with the deployed version that method only runs once when the app is built and no more. Another thing which may aid in the issues here is perhaps we need to set the `id` in the `getStaticProps` since you've set the `fallback` to blocking. This is all spitballing at this point as i'm literally learning this all in the moment haha

@narbs91 @codingwithmanny  @with-heart 
So i tried `yarn build` and it still works 😂😅 
and this is what i got

image

but it still does not work in deployment 

image

narbs91:
@moiz-lakkadkutta one thing you can try is in your `getStaticPaths` method, create and return a map of objects representing the ids from 1-8000 so that all the paths will be defined at build time (look at what the [documentation](https://nextjs.org/learn/basics/dynamic-routes/implement-getstaticpaths) is showing as an example).   Your paths object should look something like this after it's generated:

```javascript
 paths: [
    { params: { id: '1' } },
    { params: { id: '2' } },
    ...
    { params: { id: '8000' } }
  ],
```
then you'll want to return that object in the `getStaticPaths` method.  If that ends up working out we can consider just having a static json file defined in the project and read/return that in the method; however, that's just some nit picking around build time optimizations

src/pages/id/[id].tsx
```diff
@@ -158,8 +158,13 @@ export const getStaticProps = async ({ locale }: { locale: string }) => ({
 });
 
 export const getStaticPaths: GetStaticPaths = async () => {
+  var arr = [];
+  for (let i = 0; i < 8001; i++) {
+    var dic = { params: { id: i.toString() } };
+    arr.push(dic);
+  }
```
Comment (codingwithmanny) [inline src/pages/id/[id].tsx:165]:
```suggestion
  // ids should be be 1 to 8000 - hence the `i + 1`
  const arr = Array.from(Array(7999).keys()).map(i => ({ params: { id: `${i+1}` }}));
```

codingwithmanny (changes requested):

codingwithmanny:
**Good news:** `/id/1` work, great job on that.

**Still needs to be figured out:** `/fr/id/1` = 404

moiz-lakkadkutta:
> **Good news:** `/id/1` work, great job on that.
> 
> **Still needs to be figured out:** `/fr/id/1` = 404

@codingwithmanny 

Actually it works for `/en` but for any other `locale` it gives 404 

https://developer-dao-fsixi1q9w-developdao.vercel.app/en/id/98 

https://developer-dao-fsixi1q9w-developdao.vercel.app/es/id/98

I don't know what is the problem here can you have a look.

narbs91:
Just spitballing an idea here, what if you changed the path of `src/pages/id/[id].tsx ` to src/pages/[locale]/id/[id].tsx` to account for the potential locale?   I think you will probably have to modify your `getStaticPaths` implementation again to account for that change.  Building off of that idea, you could further simplify the route to `src/pages/[locale]/[id].tsx` and do a similar alteration of your `getStaticPaths` method though I'm not certain if it's the best way to approach the problem.

Not sure if anyone else has any better suggestions here

CC: @with-heart @codingwithmanny

codingwithmanny:
@narbs91 @moiz-lakkadkutta I might need to take a look at this tomorrow, sorry about the delay for the feedback, but I'll make sure to take a look at this to try and resolve it as well.

narbs91:
> Just spitballing an idea here, what if you changed the path of `src/pages/id/[id].tsx` to `src/pages/[locale]/id/[id].tsx` to account for the potential locale? I think you will probably have to modify your `getStaticPaths` implementation again to account for that change. Building off of that idea, you could further simplify the route to `src/pages/[locale]/[id].tsx` and do a similar alteration of your `getStaticPaths` method though I'm not certain if it's the best way to approach the problem.
> 
> Not sure if anyone else has any better suggestions here
> 
> CC: @with-heart @codingwithmanny

How are you doing on this feature @moiz-lakkadkutta?  Were you able to make any progress with some of the idea's around changing the path to be dynamic?

moiz-lakkadkutta:
> > Just spitballing an idea here, what if you changed the path of `src/pages/id/[id].tsx` to `src/pages/[locale]/id/[id].tsx` to account for the potential locale? I think you will probably have to modify your `getStaticPaths` implementation again to account for that change. Building off of that idea, you could further simplify the route to `src/pages/[locale]/[id].tsx` and do a similar alteration of your `getStaticPaths` method though I'm not certain if it's the best way to approach the problem.
> > Not sure if anyone else has any better suggestions here
> > CC: @with-heart @codingwithmanny
> 
> How are you doing on this feature @moiz-lakkadkutta? Were you able to make any progress with some of the idea's around changing the path to be dynamic?

No as you suggested above i  did add all the locales hardcode and it worked on development and build version on my local host but on vercel it gave 

image

@narbs91 if you can have look at the code that would be great.

narbs91:
Will do, I'll make some time tomorrow to play around with it for a bit

narbs91:
Hi @moiz-lakkadkutta i think i figured out the issue you were having with the locales. So the path should be as it was before, with `pages/id` and the file renamed `[id].js`.  Below is the code snippet that should replace you `getStaticPaths`:

```javascript
export const getStaticPaths: GetStaticPaths = async () => {
  const locales = [
    'en',
    'es',
    'fr',
    'pt',
    'pt-BR',
    'pt-PT',
    'ru',
    'zh',
    'zh-CN',
    'zh-HK',
    'zh-TW',
  ];

  const paths = [];
  for(let i = 0; i < 8000; i++) {
    for(let j = 0; j < locales.length; j++) {
      paths.push({
        params: {
          id: `${i + 1}`
        },
        locale: locales[j]
      })
    }
  }

  return {
    paths,
    fallback: false,
  };
};
```

Let me explain what's going on here: so what happens at build time of the application, is that nextjs goes and creates static pages(html and json) according to what you specify via `getStaticPaths`. In our case, there are supposed to be locales*id's of these which is 88000 with our current setup.  Now the way you have you current path setup, you had specified the routes to be generated as `[locale]/[id].js`.  For each locale, what nextjs is "trying" to do with that setup was generate the html/json for all the following `en/en/1...8000`, `fr/fr//1...8000`, and so on which isn't quite what we wanted as we were looking for was something like `en/id/1...8000`.  Now what changed from your initial implementation to what i shown above is two key things:

1. The actual looping itself, with the 8000 id's on the outside and the locales on the inside (no big deal, same number of iterations); however, take note of how the locales are now set, outside the params:
```javascript
paths.push({
        params: {
          id: `${i + 1}`
        },
        locale: locales[j]
      })
```
This is how nextjs, during build time, will know how to automagically create a [id].html and [id].json per locale that we've specified in our configs. 
2. I've set fallback to `false`.  So the purpose of fallback if set `true`(as far as i've learned) is to be a mechanism to help you render an intermediary page/component for any page that you don't tell nextjs to render at build time.  So you can imagine if we had an api with a database of a million id's, we wouldn't want to generate all these pages statically since it'll take forever to build.  Instead we'd probably want to generate a select few statically and then define a `getStaticProps` method to "fill in the gap" with the data we need to generate that new page.  For the id's we didn't generate statically, we probs show an interstitial(i.e. a fallback) until our api returned the props we needed for the page to be displayed and be cached for future use.  In our case, our id's are only ever going to be 1-8000 so we don't need to fetch anything from anywhere to define our paths.  Furthermore, won't need to define a fallback since we're telling nextjs to build all our static pages out already so that we can serve them to our users immediately.  

Even before changing your existing code, go ahead and do a `yarn build` and look at the `.next/server/pages` folder and observe what nextjs did as a result of building out all your pages.   Now, replace the code with what i've recommended at the top and run the build command again to see how things changed.

The crummy thing with this is our build times will get slower as we increate the number of locales but it's not that big of a deal at this point in time.  Please let me know if that makes sense and if you need any further clarification; however, i do think this will do the trick this time ;)

narbs91:
@with-heart @Dhaiwat10 @codingwithmanny Are you guys able to view the build logs on Vercel?  Every time I try to view them i get a 404 page

with-heart:
@narbs91 as in like https://vercel.com/developdao/developer-dao/FC85DGMFnnMu8C3oWx9BpVnqJDTv is a 404?

narbs91:
> https://vercel.com/developdao/developer-dao/FC85DGMFnnMu8C3oWx9BpVnqJDTv

yep, i get a 404 when i land on that page (signed in)

Dhaiwat10:
> @with-heart @Dhaiwat10 @codingwithmanny Are you guys able to view the build logs on Vercel?  Every time I try to view them i get a 404 page

I don't have Vercel access mate so wouldn't be able to tell you 😆

with-heart:
> > @with-heart @Dhaiwat10 @codingwithmanny Are you guys able to view the build logs on Vercel?  Every time I try to view them i get a 404 page
> 
> I don't have Vercel access mate so wouldn't be able to tell you 😆

You getting a 404 too @Dhaiwat10?

Dhaiwat10:
Yes!
image

with-heart:
@thomasmetta could we add @narbs91 and @Dhaiwat10 to the Vercel org? They're members of the site team now so they should have access imo

I'm trying to figure out how to make preview deploy inspection public. I know we can do it because Chakra [has publicly viewable deploy inspections](https://vercel.com/chakra-ui/chakra-ui/8pYFj7FHfq4u55PLLuAmre25tuUV).

moiz-lakkadkutta:
@with-heart  @narbs91 

any updates?

narbs91:
Hi @moiz-lakkadkutta sorry to leave you hanging but unfortunately i still don't have access to the build logs on vercel so I can't really tell what's going on.  Locally running a build via `yarn build` does work even though it does take a bit for it to finish.  My suspicion is that the current server may be under provisioned for all the extra pages that are being built out during the build.  I don't think it's the actual build time itself as we supposedly have a 45 min limit according to this [document](https://vercel.com/docs/concepts/limits/overview).  

Now whether or not there's a smarter way to go about things (i.e. without having to generate all possible pages for all locales) I'm not sure but I figure there should be a way.  While we wait for access to the build logs, @Dhaiwat10 I think you were involved in making the website based on next.js if i'm not mistaken, would you have any recommendations here?

narbs91:
Actually since you have access to the build logs @with-heart would you be able to take a peek tomorrow if you have some time?

narbs91:
@moiz-lakkadkutta let's try a shift to using `fallback:true` in our staticPaths method and only building the paths for just the `en` locale:

```javascript
export const getStaticPaths: GetStaticPaths = async () => {
  const paths = [];
  for(let i = 0; i < 8000; i++) {
      paths.push({
        params: {
          id: `${i + 1}`
        },
        locale: 'en'
      })
  }

  return {
    paths,
    fallback: true,
  };
};
```


And for our fallback, we'll try showing an interstitial in place of the actual NFT loading while next.js builds out the page in the background.  Just before the return statement in the `App` method, place this line of code:

```javascript
if (router.isFallback) {
    return (
      <PageLayout>
        <chakra.main>
          <VStack mx="auto" px={4} spacing={5} w="full" maxW="lg">
            <Logo w={32} h={32} />
              <CircularProgress isIndeterminate color="green.300" />
          </VStack>
        </chakra.main>
      </PageLayout>
    )
  }
``` 

So in this approach, we're telling next.js to build us only the en locale at build time and letting it figure out the rest at runtime.  While it's building the html and json in the background, we show the interstitial in the centre until it's done it's magic

narbs91:
[This](https://chakra-ui.com/docs/feedback/circular-progress) is the documentation around the interstitial from Chakra if you were curios

Dhaiwat10:
For `[id].tsx`, I switched to server-side rendering instead of static generation. Here is the build output:

image

narbs91:
Heya @marbiano, hope all is well, I was wondering if we ever hear back from that Vercel slack channel on our issues with the build?

moiz-lakkadkutta:
> Heya @marbiano, hope all is well, I was wondering if we ever hear back from that Vercel slack channel on our issues with the build?

yeah any update?

codingwithmanny:
@narbs91 any update on the support ticket?

narbs91:
> @narbs91 any update on the support ticket?

Nothing yet but I'll send a chaser today to see if there's any updates

narbs91:
Sent another chaser today as we still haven't heard back quite yet from Vercel

narbs91:
@Developer-DAO/site so Vercel got back to me today regarding the fix (and quite ironic i actually figured this out for another project yesterday lol) and what we're supposed to do is add the following line to the `next-i18next.config.js` file:

```javascript
localePath: path.resolve("./public/locales"),
```

narbs91:
I realize this PR is way behind but i think getting this is is still important to get the localization to work properly

narbs91:
Dang, maybe not as simple as that then.  I'll try and have another stab at it sometime this week (anyone else is welcome to as well)

narbs91:
Yay! This looks to be working now:

image


The issue was that the default path to the localization files was pointing to the wrong place and we needed to specify where they were (i.e in `./public/locales`).  Anyone this branch is way behind at the moment wrt to the rest of the repo, but i can work through the merge conflicts sometime this week unless you wanted to take this over the line @moiz-lakkadkutta as it was your PR

narbs91:
This was the specific resource sent to me by the Vercel support if your interested: https://github.com/isaachinman/next-i18next/issues/1448

Dhaiwat10:
Awesome @narbs91! Let's rebase & merge this?

moiz-lakkadkutta:
> Yay! This looks to be working now:
> 
> image
> 
> The issue was that the default path to the localization files was pointing to the wrong place and we needed to specify where they were (i.e in `./public/locales`). Anyone this branch is way behind at the moment wrt to the rest of the repo, but i can work through the merge conflicts sometime this week unless you wanted to take this over the line @moiz-lakkadkutta as it was your PR. Do note there are a bunch of new pages to that will need to be tested to be sure things are all working as intended

by all means go ahead

narbs91:
> Awesome @narbs91! Let's rebase & merge this?

I'll see if I have some time after work today to give it a go.  We just have to make sure things are still working as a whole since this PR is so far behind.

narbs91:
I will close this issue as of now since there is no longer any value in having this feature; however, i will make a follow up PR to add the localization change as that is something we can use in the future.  Sorry we didn't get to close out this issue fully @moiz-lakkadkutta but there will be lot's of opportunity in the future to build more cool stuff :)

PR #111, opened 2021-09-23
Added devdao-data.json as a zip

### LINKED ISSUE #101: Publish token data
I've seen a number of uses of the data related to the DevDAO token (traits, $DEV token combinations, etc.) in this project, `pixel-avatars`, and other community projects. It'd be nice to create an official package that exports this data, so all of the projects can reference the same thing rather than duplicating it all over the place. Maybe `@developer-dao/data` or `@developer-dao/token`?

codingwithmanny:
### What does it do?

Includes all 8000 ids for all tokens as a json data zipped because unzipped the file is ~23 MB

We need to give other people official data to use.

None.

### Relevant screenshots/gifs

None.

### Does it close any issues?

Closes #101


thomasmetta (approved):
nice

Issue #64, opened 2021-09-14
Gitcoin grant


thomasmetta:
Let's apply for Gitcoin grant https://gitcoin.co/grants/quickstart

thomasmetta:
There's a community and NFT Grant and i think we can apply to both https://gitcoin.co/grants/

with-heart:
Hi @thomasmetta! No pressure here, just wondering if there's an update or if there's anything we can do to help!

thomasmetta:
Hey, sorry I have been MIA this week lol. I will have something draft by tonight/tomorrow

thomasmetta:
Draft

What is DeveloperDAO
The project was started as an experiment by Nader Dabit to see what can be achieved when we bring together talented and like-minded people who are interested in learning about and building on Web3.

The DAO operates inside our discord server and over 240 active users. To gain access to this server you need to own a Devs for Revolution NFT. The contract for purchasing one can be found on etherscan.io. Our community is composed of developers from different backgrounds, we even have one of the core React team member attended our meeting

Community related notes: https://github.com/Developer-DAO/community

Initial deployment of DeveloperDAO NFT: 
https://etherscan.io/token/0x25ed58c027921e14d86380ea2646e3a1b5c55a8b

We have 2 main projects we are working on

DeveloperDAO website:
https://github.com/Developer-DAO/developer-dao

This is the public DeveloperDAO site repo. At the moment it allows members to view their genesis NFT by entering in its ID. As the community takes shape we will decide what this website becomes and build it as a collective.

There’s over 25 active developers working on sites and shipping features everyday

Derivative: https://github.com/Developer-DAO/pixel-avatars
Derivatives of the DeveloperDAO genesis NFTs; similar to how Loot has its own derivatives projects

With your support, we hope to accomplish the following:

- Service cost for discourse. Lot’s of DAO is organized discussion around proposals by using Discourse

- Gas fee for the derivative project smart contract development. Gas fees are expensive right now and we need funding in order to deploy the contract

- Bounty to reward developers who worked on features

- Build the largest community where developers can learn more about web3

- Produce educational material to onboard more web2 users and help them transition to web3

gjsyme:
In the "With your support" section -> Server cost for discourse "Lots" should be used vice "Lot's"

Dhaiwat10:
> - Bounty to reward developers who worked on features

I think we can reword this as:

- Bounties to reward content creators who have produced educational content & developers who have worked on features.

[...]

PR #125, opened 2021-09-30
Gitcoin grant proposal

thomasmetta:
### What does it do?

Turning the Gitcoin grant draft proposal https://github.com/Developer-DAO/developer-dao/issues/64 into markdown file for further use like applying for Ethereum grant. Once this is approved, I will submit to Gitcoin

### Does it close any issues?

Closes https://github.com/Developer-DAO/developer-dao/issues/64

### DIFF
--- a/doc/grant.md
+++ b/doc/grant.md
@@ -1,0 +1,59 @@
+## What is DeveloperDAO
+
+The project was started as an experiment by [Nader Dabit](twitter.com/dabit3) to
+see what can be achieved when we bring together talented and like-minded people
+who are interested in learning about and building on Web3.
+
+The DAO operates inside our discord server and have over 930+ active users. To
+gain access to this server you need to own a Devs for Revolution NFT. The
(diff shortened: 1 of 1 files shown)


gjsyme (changes requested):
Definitely should change:
- "we even have one of the core React team member attended our meeting" to correct grammar "we even have had one of the core React team members attend our meetings". To make it less react specific, "members include core team members from Dev.to, Chakra-UI, React, The Graph, and more" or something to that effect.

Other feedback (less important and my opinion based)
- I think "There are" should be used instead of "There's" for "There's over 25 active developers..."
- maybe capitalize "discourse" in "Server cost for discourse" as you do in the next sentence
- maybe remove "right now" from "Gas fees are expensive right now and..."
- use "devs" instead of "dev's" for the contraction of plural developers

narbs91:
Awesome stuff @thomasmetta !  My only nitpick would be changing `DeveloperDAO` to `Developer DAO` to match what we have on the website

PR #127, opened 2021-10-01
Log i18n key misses per language file

gjsyme:
scan_i18n.js will iterate over the source files in a given directory to
find any references to the i18n call "t('KEY')" and make a list of
unique keys from the source. It then checks all language files against
that, and notes any misses from the list of in use keys at the end of
its run.

### What does it do?
Creates the utility js file scan_i18n.js that will iterate over all source files in the repo to find the `t('KEY')` pattern and isolate keys, reduce them to a unique list, and then iterate over each language file and note any keys which are not in those files, resulting in a print to the command line of files and their missing keys or a success message. I also created a yarn script to run this as configured for our repo and a husky precommit to call that yarn script.

The need for such a script/utility/package was brought up on the website call today.

No. This was created to avoid bringing in an external package. Tradeoff: it may be frail or not quite properly tuned (with the regex, for example) which is why I am creating this as a draft

### Relevant screenshots/gifs
image

### Does it close any issues?
No. Unless I/someone else opens one before I make this a real PR

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -2,3 +2,4 @@
 . "$(dirname "$0")/_/husky.sh"
 
 npx lint-staged
+yarn scan-i18n
--- a/package.json
+++ b/package.json
@@ -44,7 +44,8 @@
     "test": "jest",
     "test:watch": "jest --watch --detectOpenHandles",
     "generate-types": "typechain --target ethers-v5 src/artifacts/*.json && prettier --write \"types/ethers-contracts/**/*.ts\"",
-    "postinstall": "yarn generate-types"
+    "postinstall": "yarn generate-types",
+    "scan-i18n": "node scan_i18n.js ./src ./public/locales"
   },
   "eslintConfig": {
     "extends": [
--- a/scan_i18n.js
+++ b/scan_i18n.js
@@ -1,0 +1,89 @@
+const fs = require('fs');
+const path = require('path');
+
+// reads the arguments at the command line:
+// index 2 is the source code directory
+// index 3 is the language directory
+// iterates over the source directory, recursively, to find any call to t('YOUR_KEY_HERE')
+// once all keys have been found, it iterates over the langage directory and checks each translation file to see if it has the i18n keys expected
+// if any don't, then they are noted
+// after iterating over all language files, the resulting misses are printed to the command line
+// but if everything is translated, then it prints a success message instead.
+
+if (process.argv.length !== 4) {
+  console.error(
+    `unexpected arguments; expected 2, saw ${process.argv.length - 2}`,
+  );
+  return;
+}
+console.log(
+  'scanning internationalization keys in',
+  process.argv[2],
+  'against language files in',
+  process.argv[3],
+);
+
+function getAllFiles(dirPath, fileArray) {
+  fs.readdirSync(dirPath).forEach(function (file) {
+    let filePath = path.join(dirPath, file);
+    let stat = fs.statSync(filePath);
+    if (stat.isDirectory()) {
+      getAllFiles(filePath, fileArray);
+    } else {
+      fileArray.push(filePath);
+    }
+  });
+  return fileArray;
+}
+
+const fileArray = getAllFiles(process.argv[2], []);
+const languageFiles = getAllFiles(process.argv[3], []);
+
+const regex = /(\s|\{)t\(\'(\w*)\'/;
+const keys = [];
+const rejects = [];
+fileArray.forEach((filePath) => {
+  const fileContents = fs.readFileSync(filePath, {
+    encoding: 'utf-8',
+    flag: 'r',
+  });
+  fileContents.split('\n').forEach((textLine) => {
+    const result = regex.exec(textLine);
+    if (result) {
+      keys.push(result[2]);
+    }
+  });
+});
+
+// now reduce to uniques
+const uniqueKeys = [...new Set(keys)];
+const missTracker = {};
+
+// and now we can check our translations for these keys
+languageFiles.forEach((filePath) => {
+  const fileContents = fs.readFileSync(filePath, {
+    encoding: 'utf-8',
+    flag: 'r',
+  });
+  // turn that filecontent back into an object and get its keys
+  const contentObject = JSON.parse(fileContents);
+  uniqueKeys.forEach((key) => {
+    if (!contentObject[key]) {
+      if (!missTracker[filePath]) {
+        missTracker[filePath] = [];
+      }
+      missTracker[filePath].push(key);
+    }
+  });
+});
+
+if (Object.keys(missTracker).length > 0) {
+  Object.keys(missTracker).forEach((fileName) => {
+    // red: \x1b[31m%s\x1b[0m
+    // yellow: \x1b[33m%s\x1b[0m
+    console.error('\x1b[31m%s\x1b[0m', `Missing keys in ./${fileName}:`);
+    console.log('\x1b[33m%s\x1b[0m', `[${missTracker[fileName].join(', ')}]\n`);
+  });
+} else {
+  console.log('Success! i18n scan found no missing keys');
+}


Dhaiwat10:
Very impressive man, neat idea! I would like to propose a slightly different approach, though.

When someone commits their work which involves typing some text, more often than not that person can only add the translations for English and maybe another language they speak. I don't think there's anyone who can speak all of the languages we support. 😅

What I propose is: instead of running this check when committing, we run this check for PRs. Or better yet, run this script on push/merge & create issues for missing translations like we are manually doing atm.

narbs91:
Awesome stuff @gjsyme, thanks for taking the initiative on this!  I would say scanning the `src/` directory is probably not a scalable solution as the number of files in the project increase like we think it will.  Plus from my initial peek at the code, it  looks like it'll scan all the files in the `src` directory which more then we probs need.  What I would suggest as a simplification to the process instead is to change the implementation to take `public/en/common.json` as the "master" translation file and do a key comparison between it and the other locale files.  For each key missing per translation file, we report it just as you're doing.  I'm also wondering if we could take this one step further and perhaps look to create a json comparator where we take two json files and compare them for equality (as we did want to also compare the package.json file as well if i'm not mistaken).  But that may be out of scope for now, as the missing translations are the more pressing issue.

Also agree with @Dhaiwat10's recommendation of running the script on merge (if that's possible) since the scan could potentially slow down the deployments/checks as the files start to grow.

gjsyme:
I'm looking at the Action option. I will keep it scanning source though as we miss a lot of value if someone hardcodes in a string and we don't pick it up. I've gotten the action to not-fail, but wasn't able to get it to actually do anything this far.

kavimaluskam:
Impressed by all your quick ideas and solutions ❤️
I think suggestion by @Dhaiwat10 is cool. 
Probably we can scan/compare the file changes when PR merge and create Issue only when new language key is added.

e.g. https://dev.to/scienta/get-changed-files-in-github-actions-1p36

gjsyme:
Just an update (as this will still eventually be a PR on this repo once the Action is written): I'm writing it as an Action (which will then be its own repo). Based on the dev.to link, I'm looking at feeding it a list of i18n keys based upon git grep (so that it can get the t('KEY') calls without iterating on files in src). I will then use that output as an input to the javascript Action to check each Key file.

I still think that the value of this Action is to per-PR point out those translations that have not been made - ideally this will be then picked up as appropriate in review. There may be a way to get an Action to compare a missing file/key pair against open i18n tickets, but I think that might be a bit excessive for the time being.

gjsyme:
Alright - sorry for the spam deploys but what worked with Act didn't quite work with the actual github actions. But now we're here, and the steps that I've added to PR run successfully and add warnings to the Checks on missed tokens. Unfortunately, that's only visible if you dive into the PR checks ("show all checks") then open up the jsrun step.
This relies on an external action repo (currently under my own GH, but we could add it to D_D with a working v1.0). Let me know what you think.

Dhaiwat10:
@gjsyme I can't see the jsrun step mate. Am I looking in the wrong place?

image

Dhaiwat10:
Btw, if we can make this work, we can open-source this as a re-usable action/package 👀

kavimaluskam:
@Dhaiwat10 thought the Same. The issue we are trying to solve should be a general i18n issue.❤️

kavimaluskam:
> @gjsyme I can't see the jsrun step mate. Am I looking in the wrong place?
> 
> 
> 
> image
> 
> 

@gjsyme I am in the exact same
Situation 😭

By the way attach your external repo here: https://github.com/gjsyme/i18n-tag-audit-action ❤️

Dhaiwat10:
I guess it only makes sense that we cannot see it here since the changes to the workflow file haven't been merged.

gjsyme:
PR workflow/build -> show details (opens another page) -> will have a list of steps -> jsrun

Dhaiwat10:
Oh, I see!

Dhaiwat10:
This is great progress. I'm gonna look into your code. There must be a way to automatically create issues for missing keys.

Dhaiwat10:
@gjsyme my man the code in the actions repo looks like it's the output after compilation. Am I right in assuming that?

Dhaiwat10:
This a great discussion on how to run a workflow when a PR is merged: https://github.community/t/trigger-workflow-only-on-pull-request-merge/17359/2

kavimaluskam:
To create an issue on github action, probably we can simply use / check the code some existing project.

e.g. :https://github.com/JasonEtco/create-an-issue

.github/workflows/pr.yml
```diff
@@ -21,6 +21,19 @@ jobs:
...
 
+      # block for checking i18n tags
+      - name: Grep translation calls
+        id: greps
+        run: echo "::set-output name=matches::$(git grep -E -o "(\s|\{)t\('(\w*)'" | tr '\n' ', ')"
+      - name: saver
+        id: saver
+        run: echo "::set-output name=output1::${{ steps.greps.outputs.matches }}"
+      - name: jsrun
+        uses: gjsyme/i18n-tag-audit-action@v1.26
```
with-heart, line 32: I think it'd be better if [we kept this action local](https://github.community/t/path-to-action-in-the-same-repository-as-workflow/16952/2) for now. That would let us dogfood it in our own repo so we can explore the space of what's most useful in our translation workflows, without losing the context of what the action does and speed of changes to it because it lives somewhere else. Does that make sense?

with-heart (changes requested):

with-heart:
Really loving all of the discussion and energy in this PR! Especially for publishing a package!

Just a few thoughts:

- I think the purpose of the action might be a little off with this approach.

I think the solution of using an action to manage this is the correct one, so it's excellent that it was proposed and that @gjsyme is exploring that! Growing our collective understanding of actions is going to be very useful to us as an org.

However, I think we're still missing a piece of the puzzle here. As @Dhaiwat10 said previously:

> When someone commits their work which involves typing some text, more often than not that person can only add the translations for English and maybe another language they speak. I don't think there's anyone who can speak all of the languages we support. 😅

If this is true, then reporting missing translations on each PR means we're reporting information that is likely unactionable to anyone involved in the PR (unless the relevant translator happens to review it). It's solid information to have available, just not in that context!

I like the suggestion for creating issues. I think that could be a better starting point because it at least allows us to target that information in a more specific way: 

- we could create a `site-translators` team on GitHub that we automatically assign to each generated issue (and then add translators to it, of course)
- we have a `translation` label that we can automatically add to each issue as well
- automatically tag the PR the issue was spawned from in the body text so they're linked together (someone looking at the merged PR would be able to follow along with follow-up translation work)

with-heart:
omfg I was editing that last comment and had a bunch more stuff and somehow lost it all. ugh

.github/workflows/pr.yml
gjsyme: Reasonable. Zero experience with actions before so I had just looked at how to build one and got going. And there's a reason that the external action is on v1.26...

gjsyme:
2 things: remove KEY and YOUR_KEY_HERE from arrays, determine why GH creation request fails "Resource not accessible by itegration"

.github/actions/i18n-tag-audit-action/node_modules/.bin/ncc
```diff
@@ -0,0 +1 @@
+../@vercel/ncc/dist/ncc/cli.js
```
narbs91, line 1: Do we need to include the node_modules in this package for the action to work?  Can't we specify a step in the script to first run the npm install and then run the script?

.github/actions/i18n-tag-audit-action/index.js
```diff
@@ -0,0 +1,102 @@
...
+    Object.keys(missTracker).forEach(async (fileName) => {
+      core.warning(
+        `Missing keys in ./${fileName}:\n [${missTracker[fileName].join(
+          ', ',
+        )}]`,
+      );
+      try {
+        const newIssue = await octokit.rest.issues.create({
+          ...github.context.repo,
+          title: `i18n misses in ${fileName}`,
```
narbs91, line 88: [nit pick] Maybe we can change this to `i18n entries missing in ${fileName}`?

.github/actions/i18n-tag-audit-action/index.js
```diff
@@ -0,0 +1,102 @@
...
+          body: `The following i18n keys miss (have no value) in ${fileName}:\n${missTracker[
```
narbs91, line 89: [nit pick] Maybe we can change this to `The following i18n keys are missing  (have no value) in ...`

.github/actions/i18n-tag-audit-action/node_modules/.bin/ncc
gjsyme: No idea. https://docs.github.com/en/actions/creating-actions/creating-a-javascript-action#commit-tag-and-push-your-action-to-github makes it sound like you do need to but I've got zero action experience / knowledge beyond what I've done here.

Dhaiwat10:
@gjsyme I was thinking about creating a repo for this action inside the Developer DAO org, and continue working on refining this there? I think that'd pave the way for potential collaboration with people from the DAO who already have some experience with Actions. Let me know what you think.

with-heart:
I was trying to play around with actions this morning to push this along but of course GH has an actions outage right now so my experiment won't run 😅 

One thing I did notice is that we'll want to run this action in a new workflow that targets `push` for `main`. As it stands, this will attempt to create "missing translation" issues every time a PR is opened or updated.

gjsyme:
I was able to move the action out to its own push.yml and remove the node_modules; this tests as well as it can given the permissions that fail on local runs (using the [act](https://github.com/nektos/act) tool). I looked into testing against my own user token, but I didn't see issue management as an option in the creator. Additionally, forked repo github contexts are read only for issues: depending on our organization settings, it may be necessary to explicitly set permission for the github token ref: https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/

.github/workflows/push.yml
```diff
@@ -0,0 +1,38 @@
...
+        uses: actions/setup-node@v2
+        with:
+          node-version: 14.x
+
+      - name: Install yarn
+        run: npm install -g yarn
+
+      # make sure we have the GH dependencies
+      - name: Install dependencies
+        run: cd .github/actions/i18n-tag-audit-action && yarn install --frozen-lockfile
```
with-heart, line 26: If we add `.github/actions/*` as a workspace, `yarn install` will automatically install the packages in the right place: https://github.com/with-heart/example-action/blob/42b0a867667f59d3df09dc81e60203823186763a/package.json#L22-L24

gjsyme:
I've timed it perfectly with @with-heart hitting the road again... but this is now finally a real PR. In theory (can't test due to forks not having rights to manage issues on main) creates 1 issue per push to track all missing i18n keys (calls to get an i18n string that don't have matching values in the json files).

gjsyme:
Alright. What a long strange trip this has been. Months later, we're back to the husky pre-commit hook and package.json script to run the scan_i18n.js script, per the last website team call.

narbs91 (approved):
LGTM!  I'll let @with-heart have the final say on this one though

PR #158, opened 2021-10-31
Show available token list on mint page

DennisKo:
### What does it do?

It shows available Tokens in a grid. Once clicked on a token it opens the mint-box with the token id pre-populated.

To find out which tokens are available I get all "Transfer" events and then check those results against a list with n=8000 (Inspired by `'https://developerdao.vercel.app/'`)

nope

### Relevant screenshots/gifs
tokenlist3

### Does it close any issues?
Should I open one first?


src/hooks/useAvailableTokens.tsx
```diff
@@ -0,0 +1,76 @@
...
+}
+
+export function useAvailableTokens() {
+  const [tokens, setTokens] = useState<string[]>([]);
+  const [isFetchingTokens, setIsFetchingTokens] = useState(false);
+
+  useEffect(() => {
+    const fetchTokens = async () => {
+      try {
+        const provider = new FallbackProvider([
```
DennisKo, line 36: Cant use defaultProvider here as it has problems with large responses: -> https://github.com/ethers-io/ethers.js/discussions/2248

src/pages/mint.tsx
```diff
@@ -1,17 +1,25 @@
...
 
 const Mint = () => {
   const { t } = useTranslation();
+  const [selectedToken, setSelectedToken] = useState<string | undefined>();
   return (
     <>
       <PageLayout>
         <chakra.main>
-          <DirectMintBox />
+          {selectedToken && (
```
DennisKo, line 16: It might be better to have that as a subroute and then properly link to that and pass a url param

with-heart:
I think it would make sense to render each available token individually if we were displaying some additional info about each token (like its traits or image or something), but since we're not, I feel like it would make more sense to render this as a `select`. Too many buttons make for a lot of visual noise (especially with 3k+ tokens still available) for no real benefit to the user.

with-heart:
I do like the idea of asking for them to pick a token first and then asking to connect the wallet. I think that's a nice improvement in the flow! Get them engaged first, then move on to business.

DennisKo:
> Too many buttons make for a lot of visual noise (especially with 3k+ tokens still available) for no real benefit to the user.

Yep at the moment this view does not add anything useful other than showing the available token. I would think the next iterations would show additional info about the a single token , the meta info for sure (onHover even?). Maybe even the images from the derivates project? I am envisioning the end product to be a grid view like opensea (https://opensea.io/collection/cryptopunks). Then a select would not be the best choice.

narbs91:
Perhaps another way we could look towards is having a single available token show up with the traits and a shuffle button to swap in another one if the user doesn't like the one they were presented.   It's tough to come up with something that works for thousands of items but perhaps i do think a "in-between" selection page with the actual NFT images (like we show on the current landing page) would probs be a good  north star vision for this concept.  Great stuff overall 👌

PR #159, opened 2021-11-01
Redesign landing page

Nazeeh21:
### What does it do?
Developed the **Developer_Dao** website.

### Any helpful background information?
This PR contains the newly developed **Developer Dao** website according to the [figma](https://www.figma.com/file/NuGyq4CdJaCFAt6oubMcVv/D_D?node-id=7%3A2) design. The newly developed _landing page_ can be found on `/landing-page` route. (This setup is chosen to avoid merge conflicts on the PR).

This is draft PR, so everyone is welcome to give their suggestion. The footer of the website is remained to be redeveloped according to the new design, we'll add this in the upcoming commits.

### Any new dependencies? Why were they added?
marquee library to show the tweets

### Relevant screenshots/gifs
image

### Does it close any issues?
This PR targets fixing issue #114

Closes #114

PR #160, opened 2021-11-02
Add hook to get the DEV NFT supply

embiem:
### What does it do?

Adds a hook into a new src/hooks/ folder that fetches stats about the DEV NFT supply. Can be used in the PR #159 to retrieve the "seats left" stat, but also just to improve UX on the mint page.

Also adds "We have 2909 tokens left" text on /mint page (see screenshot).

After seeing the landing-page redesign in #159, I wanted to contribute to it by writing this hook. If necessary, I can also tackle a similar hook to retrieve Discord stats.

I'd like to add 2 new dev dependencies:

[@testing-library/react-hooks](https://react-hooks-testing-library.com/), to properly test the hook.

[typechain](https://github.com/dethcrypto/TypeChain), to generate types for smart contracts. Improves DX because of code completion & safety because of type checking on smart contract function calls.

### Relevant screenshots/gifs

image

### Does it close any issues?

Nope.


src/hooks/useDevNFTSupply.ts
```diff
@@ -0,0 +1,90 @@
...
+        lockedTokenId <= MAX_SUPPLY;
+        lockedTokenId++
+      ) {
+        requests.push(
+          contract.functions.ownerOf(lockedTokenId).then(
+            // Return the owner's address
+            (addressResponse) => addressResponse[0],
+            // Return the zero-address if no owner was found
+            () => ethers.constants.AddressZero,
+          ),
```
embiem, line 56: This will create a lot of requests, but it's the only way I found to do this. 

web3 library has [BatchRequests](https://web3js.readthedocs.io/en/v1.2.11/web3-eth.html#batchrequest), which would lower the number of requests but increase the overall execution time as they're called sequentially instead.

Anyways, if anyone knows of a better way to do it, please let me know :-)

embiem:
I just pushed my latest changes. I'll stop now as it is getting bigger than I initially wanted it to. The original PR description has been edited to reflect the current changes.

In this PR, I now also included a suggestion to use typechain. This adds types that are generated via CLI (see the `postinstall` script), are git-ignored and can be used to get a typed ethers contract object. An example where types like this are useful (apart from auto-completion): changing `contract.totalSupply()` to `contract.functions.totalSupply()` will change the `BigNumber` response to `Array<BigNumer>`. Without types, an error like this could sneak through all the way to runtime.

I'd love to get your feedback on this suggestion.

src/hooks/useDevNFTSupply.ts
```diff
@@ -0,0 +1,84 @@
...
+    );
+
+    // Fetches the count of minted DEV NFTs
+    const fetchTotalSupply = async () => {
+      const totalSupply: number = (await contract.totalSupply()).toNumber();
+      setTotalSupply(totalSupply);
+    };
+
+    // Counts how many tokens of the locked supply have been minted
+    const fetchLockedSupply = async () => {
```
narbs91, line 36: If i understood the intent of this function it seems like we're fetching the locked supply so we can calculate the remaining tokens available for mint.  Since we know for a fact that there are 8000 total tokens, 3 reserved for the owner and not counting the 0 address, can we just not just do:  `availToMint = 8000 - 3 -1 - totalSupply`?  As totalSupply is the count of the currently minted NFT's

narbs91: Now the only thing i'm not super clear about is if totalSupply is deduped per address and thus represents the number of unique owners

narbs91: Just did the math 8000-222-1- 4887 = 2,889 which is the actual number of tokens left; however the UI looks like it 's counting the zero address to in it's calculations as it's showing 2890 when i landed on it

narbs91: oh nvm, just re-read and this is for the locked supply (DOH!) but i suppose this raises another thought around does the end user actually need to know this information?  Like as a person who wants to join the DAO I only really care if there's tokens available for me to mint (i.e the public supply) and not so much the tokens only available to the owner.

.gitignore
```diff
@@ -22,4 +22,6 @@ npm-debug.log*
 yarn-debug.log*
 yarn-error.log*
 
-.next
\ No newline at end of file
+.next
+
+/types/ethers-contracts
```
with-heart, line 27: Would we not want to check these in? If the types are used throughout the codebase, they should remain a part of it.

src/hooks/useDevNFTSupply.ts
```diff
@@ -0,0 +1,84 @@
...
+
+      setLockedSupply(countMinted);
+    };
+
+    fetchTotalSupply();
+    fetchLockedSupply();
+  }, []);
+
+  return {
+    loading: totalSupply === -1 || lockedSupply === -1,
```
with-heart, line 70: This might just be personal preference here but I'd like to suggest that we rename this to `isLoading`. That naming follows a convention used by a number of other libraries ([Chakra UI](https://chakra-ui.com/docs/form/button#button-loading-state), [`react-query`](https://react-query.tanstack.com/guides/queries)) and prefixing variables with `is`/`has`/`are` makes it obvious at a glance that the variable is a `boolean` without needing to check the type signature.

with-heart (changes requested):
I love the thinking here! A library of general-purpose hooks make for a solid foundation for an app.

src/hooks/useDevNFTSupply.ts
embiem: Yes I agree with you. When getting the totalSupply() via the one function call on the contract we get the supply of public & locked tokens, so even if we want to show the correct number of available public tokens we'd still need to know what portion of the total supply is actually locked.

.gitignore
embiem: Fair point. I usually don't include any generated files in git, so that's why I put it on the git-ignore list initially.

src/hooks/useDevNFTSupply.ts
embiem: Agreed! I'll change it.

src/hooks/useDevNFTSupply.ts
embiem: I just saw that DennisKo used `contract.queryFilter` in #158 to get all transactions and retrieve all available tokens this way. Not sure whether this might be the better approach to fetch how many of the locked tokenIds were already minted.

Alternatively I was thinking that this must be the kind of info that The Graph is trying to index. Should be more efficient to actually index this info instead of always calling the contract's read functions. I have 0 experience with The Graph yet, though.

embiem: > Like as a person who wants to join the DAO I only really care if there's tokens available for me to mint (i.e the public supply) and not so much the tokens only available to the owner.

That's the number I show on the /mint page. It's returned as the `remainingPublicSupply`, which is `PUBLIC_MAX_SUPPLY - (totalSupply - lockedSupply)`.

narbs91: It might worthwhile to poke around using the `contract.queryFilter` to see if we can cut down on the number of network calls (haven't had any experience using this myself so can't say for sure it'll work).  With regards to The Graph, i think you're right that it probs has the info indexed already but querying against a subgraph costs money so i don't know if it's a viable solution in our current stage of the DAO

with-heart (approved):

embiem:
Thanks for approving @with-heart! I do have one more thing in the pipe though, based on @narbs91's comment to try out `contract.queryFilter` and my Discord message where everyone is saying we should display the amount of unique token holder addresses instead of the total supply.

embiem:
I just added the final commit, please let me know if there's any more feedback on this.

src/hooks/useDevNFTSupply.ts
```diff
@@ -0,0 +1,100 @@
...
+    lockedSupply,
+
+    // This is the amount of publicly minted DEV NFTs, which is open to mint by anyone (TokenIDs 1-7777, incl.)
+    publicSupply: totalSupply - lockedSupply,
+
+    // Total number of possible DEV NFT. This is a constant that cannot be changed
+    maxSupply: MAX_SUPPLY,
+
+    // Some calculations to make these values easily accessible for display in UI
+    remainingTotalSupply: MAX_SUPPLY - totalSupply,
```
narbs91, line 96: [nit pick] do we need to return `remainingTotalSupply` and `remainingLockedSupply` if we're not using them upstream?  Or did you leave here incase we wanted to use them in the future?

narbs91 (approved):
LGTM!  Made a minor comment on some of the return elements in useDevNFTSupply but other then that it looks great, awesome job

src/hooks/useDevNFTSupply.ts
embiem: Yes I left them for any peobable future use, but I can see how this might add bloat when we do this a lot. So I'd be happy to remove as well.

narbs91: Nah, we can keep them here for now all good

narbs91:
@allcontributors please add @embiem  for code

PR #161, opened 2021-11-03
Add a Resources page

gjsyme:
### What does it do?
Pulls data from the Airtable resource base to populate a list of cards that people can view right from the developerdao.com website.

This is identified as our lowest hanging of educational fruits and something that we want to get out rapidly.

`airtable`; this provides support for the Airtable API so that we can fetch the data.

### Relevant screenshots/gifs
Deferred until someone can take a swipe and cleaning up the UI.

### Does it close any issues?
No

### New keys have been created
New i18n keys will require translations for all languages other than english


src/utils/AirtableResourceClasses.ts
```diff
@@ -0,0 +1,48 @@
+// simplified airtable objects that work for our purposes
```
with-heart, line 1: Could we rename this to `src/types/airtable.ts`? These aren't really `utils` 😁

src/pages/resources.tsx
```diff
@@ -0,0 +1,287 @@
...
+} from '../utils/DeveloperDaoConstants';
+import {
+  Author,
+  Blockchain,
+  Category,
+  Resource,
+  Tag,
+} from '../utils/AirtableResourceClasses';
+
+function KnowledgeBase(props: { airtable: Airtable }) {
```
with-heart, line 33: `props` aren't actually used!

```suggestion
function KnowledgeBase() {
```

src/pages/resources.tsx
```diff
@@ -0,0 +1,287 @@
...
+  const { t } = useTranslation();
+  const [airbase, setAirbase] = useState('Resources');
+  const [airtable, setAirtable] = useState<Airtable>();
+  const [resourceBaseList, setResourceBaseList] = useState<Resource[]>([]);
+  const [showAll, setShowAll] = useState(false);
+  const toast = useToast();
+
+  useEffect(() => {
```
with-heart, line 41: imo we could add a lot of clarity here by separating the airtable stuff out into a hook. That way someone wanting to understand what the component is rendering can just look at its body (which will be quite short) and if they want to dig deeper into the complexity of how we're managing the airtable data, they can then read through the hook. 

```tsx
const useKnowledgeBase = () => {
  // `useTranslation`, `useToast`, `useState`, and `useEffect` calls
  return viewList
}

const KnowledgeBase = () => {
  const knowledge = useKnowledgeBase()
  return (
    {knowledge.map(item => {
      // cool stuff here or whatever
    }}
  )
}
```

with-heart (changes requested):

with-heart:
This is gonna be so nice to have!

src/components/ResourceCard/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+      case 'Article':
+        return t('readArticle');
+      case 'Course':
+        return t('checkCourse');
+      case 'Book':
+        return t('checkBook');
+      default:
+        return t('unknownMediaText');
+    }
+  };
```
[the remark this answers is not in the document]
gjsyme: React objected to me returning a string as a component so I have taken the middle path of making a MediaButton that builds the entire button (including a MediaIcon as componentized above and the text being determined by a method in the MediaButton).

src/components/ResourceCard/index.tsx
```diff
@@ -22,61 +22,85 @@ import {
...
+    case 'Course':
+      return <Icon as={FaChalkboard} />;
+    case 'Book':
+      return <Icon as={FaBook} />;
+    default:
+      return <Icon as={FaQuestion} />;
+  }
+};
+
+// const MediaText = ({mediaType}: {mediaType: string}) => {
```
narbs91, line 83: [nit pick] Did you mean to leave this commented code here?

src/components/ResourceCard/index.tsx
```diff
@@ -91,14 +115,30 @@ function KnowledgeCard(props: { data: Resource }) {
...
+  const { t } = useTranslation();
+
+  const kbRecord = props.data.fields;
+
   return (
     <WrapItem maxW={{ base: '100%', md: '800px' }}>
       <VStack>
         <Heading>{kbRecord.Title}</Heading>
-        <Box>{getAuthor(kbRecord.extendedAuthors)}</Box>
+        {/* <Box>{getAuthor(kbRecord.extendedAuthors)}</Box> */}
```
narbs91, line 137: [nit pick] ditto here around commented code

src/pages/resources.tsx
```diff
@@ -175,19 +169,13 @@ function KnowledgeBase(props: { airtable: Airtable }) {
...
-    }
-    if (categoryList.length == 0) {
-      return;
-    }
-    if (resourceList === undefined || resourceList.length === 0) {
-      return;
-    }
-    if (tagList.length == 0) {
+    if (
+      !authorList.length ||
```
narbs91, line 173: might be cleaner to put this in it's own private function?

src/pages/resources.tsx
```diff
@@ -243,16 +231,10 @@ function KnowledgeBase(props: { airtable: Airtable }) {
...
-      return resourceBaseList.filter((item) => item.fields.Curated);
-    }
-  };
-  const getViewLength = () => {
-    return getViewList().length;
-  };
+
+  const viewList = showAll
+    ? resourceBaseList
+    : resourceBaseList.filter((item) => item.fields.Curated);
```
narbs91, line 237: Can item.fields ever be null || undefined?  If so maybe we can add a check here around it (i think you can do `item.fields?.Curated` if i'm not mistaken"

src/components/ResourceCard/index.tsx
```diff
@@ -0,0 +1,163 @@
...
+          <LevelBadge levelName={kbRecord.Level} />
+        </HStack>
+        <Text>{t('tags')}</Text>
+        <Stack direction="row">
+          {kbRecord.Tags?.map((tag: string, index: number) => (
+            <Badge key={`tag-${index}`}>{tag}</Badge>
+          ))}
+        </Stack>
+        <HStack>
+          <Link href={kbRecord.Source} _hover={undefined}>
```
narbs91, line 150: Just curious, why do we have the `_hover={undefined}` here?  Mostly around the `undefined` lol

src/utils/DeveloperDaoConstants.ts
```diff
@@ -9,6 +9,9 @@ export const MAINNET_NETWORK_ID = 1;
 export const TOKEN_FINDER_URL = 'https://developerdao.vercel.app/';
 export const ETHERSCAN_TX_URL = 'https://etherscan.io/tx/';
 
+// airtable resources managed by @kempsterrrr
+export const AIRTABLE_READONLY_KEY = 'keyDaaDlZelNrerXQ';
```
narbs91, line 13: Is this key something that has to be paid for or has a rate limit applied to?  I'd push for us to have this as an env variable as we should strive not to expose apiKeys or what not in a public repo.   I can add an env variable in vercel for this entry:  NEXT_PUBLIC_AIRTABLE_READONLY_KEY and then you can access it like `process.env.NEXT_PUBLIC_AIRTABLE_READONLY_KEY`

narbs91: I've added this env variable in vercel for the readonly key

src/utils/DeveloperDaoConstants.ts
```diff
@@ -9,6 +9,9 @@ export const MAINNET_NETWORK_ID = 1;
...
+export const AIRTABLE_RESOURCE_BASE = 'appPP6MpO5hrfQwqI';
```
narbs91, line 14: Not sure what this is for but we can take the same approach if this is something that shouldn't be out in public

narbs91: Eh i added this one too to vercel:  NEXT_PUBLIC_AIRTABLE_RESOURCE_BASE

src/components/ResourceCard/index.tsx
gjsyme: I did when I did it and then forget to clean it up afterwards. Thanks!

src/pages/resources.tsx
gjsyme: Looked into this as the assumption of fields being non-falsey is a lot of places. The airtable sdk has ```setRawJson(rawJson: RecordJson): void {
        this._rawJson = rawJson;
        this.fields = (this._rawJson && this._rawJson.fields) || {};
    }```
    which seems to be the way they update a Record's data -> we'll always at least have a fields object.

src/components/ResourceCard/index.tsx
gjsyme: Seems useless. Was in the sample I made on links-as-buttons in chakra but doesn't seem to change behavior.

src/utils/DeveloperDaoConstants.ts
gjsyme: Perhaps worth coordinating with @kempsterrrr and then replacing it behind the scenes. 
There is a rate limit: 
```RATE LIMITS
The API is limited to 5 requests per second per base. If you exceed this rate, you will receive a 429 status code and will need to wait 30 seconds before subsequent requests will succeed.

The official JavaScript client has built-in retry logic.

If you anticipate a higher read volume, we recommend using a caching proxy. This rate limit is the same for all plans and increased limits are not currently available.```

File: src/utils/DeveloperDaoConstants.ts (hunk 11 above)
Comment (gjsyme) [reply src/utils/DeveloperDaoConstants.ts:14]: What's the solution for production using the vercel-set value and non-production using the locally defined variable?

**Comment by gjsyme** [comment]
Where we want this to be a public good, I think we need to figure out the caching proxy they refer to as people would be loading that every day. I'm going to say that's a "new issue" kind of problem though.

File: src/utils/DeveloperDaoConstants.ts (hunk 11 above)
Comment (narbs91) [reply src/utils/DeveloperDaoConstants.ts:14]: So production would be covered by the variable being set in vercel but for non-prod (basically locally) the solution is to add a `.env` file to the project root and add the key/value there.  This is something that we won't want people to commit and it's already in the `.gitignore` if i'm not mistaken.  We can push up a sample.env though that has the keys but not the values and have people ask for the values from one of us.  Though the issue then becomes who do we give the secret too and who do we not which is probs a process we have to come up with

File: src/utils/DeveloperDaoConstants.ts (hunk 10 above)
Comment (narbs91) [reply src/utils/DeveloperDaoConstants.ts:13]: ah i see, i guess it really depends on the traffic we get on the page; however, if the content is fairly static in nature we could always look to introduce some caching on our end (separate issue obviously)

File: src/pages/resources.tsx (hunk 7 above)
Comment (gjsyme) [reply src/pages/resources.tsx]: I don't agree where it's a one-time block. I may be wrong, but it seems like just moving that block of or statements around if we make it a function.

**Comment by gjsyme** [comment]
I opened an issue for the caching proxy #166 as it will resolve the question of exposing keys anyway (proxy would have to have them; queries will go there vice direct to airtable).

File: src/utils/DeveloperDaoConstants.ts (hunk 11 above)
Comment (with-heart) [reply src/utils/DeveloperDaoConstants.ts:14]: `.env.local` instead of `.env` :)

File: src/utils/DeveloperDaoConstants.ts (hunk 10 above)
Comment (gjsyme) [reply src/utils/DeveloperDaoConstants.ts:13]: @narbs91 if you can update the constants to be server side, then we're good to go on this. the work by @Neumand on my fork lets us make the call to airtable on the server side (and restrict how often it will refresh) so we dont' even need NEXT_PUBLIC_

File: src/utils/DeveloperDaoConstants.ts (hunk 11 above)
Comment (gjsyme) [reply src/utils/DeveloperDaoConstants.ts:14]: see above - server side constants work (same names, just no 'NEXT_PUBLIC_')

File: src/utils/DeveloperDaoConstants.ts (hunk 10 above)
Comment (narbs91) [reply src/utils/DeveloperDaoConstants.ts:13]: Cool beans, i'll change them up to remove the NEXT_PUBLIC prefix

File: src/lib/airtable.ts [hunk 12]
```diff
@@ -0,0 +1,97 @@
...
+
+  authors.forEach((item) => {
+    authorMap[item.id] = {
+      name: item.fields.Name,
+      dev: item.fields['Developer DAO Member'],
+    };
+  });
+
+  blockchains.forEach((item) => {
+    blockchainMap[item.id] = item.fields.Name;
```
narbs91, line 39: is it possible for `item.fields` to be null or undefined?

src/lib/airtable.ts
```diff
@@ -0,0 +1,97 @@
...
+  });
+
+  tags.forEach((item) => {
+    tagMap[item.id] = item.fields.Name;
+  });
+
+  return resources.map((resource: Resource) => {
+    const newFields = {
+      ...resource.fields,
+      ...(resource.fields.Author?.length < 0 && {
```
narbs91, line 53: Same question here around the possibility of `resource.fields` being null or undefined

src/pages/resources.tsx
```diff
@@ -0,0 +1,58 @@
...
+    <PageLayout>
+      <chakra.main>
+        <HStack>
+          <Text>{t('curatedResources')}</Text>
+          <Switch isChecked={showAll} onChange={updateListFilter} />
+          <Text>{t('allResources')}</Text>
+        </HStack>
+        <Box>
+          {viewList.length === 0 ? (
+            <Text>No Records</Text>
```
narbs91, line 32: Looks like we missed this piece of text getting added to the translation files

src/utils/useAirtableResources.ts
```diff
@@ -0,0 +1,216 @@
...
+export const useAirtableResources = () => {
+  const { t } = useTranslation();
+  const [airbase, setAirbase] = useState('Resources');
+  const [airtable, setAirtable] = useState<Airtable>();
+  const [resourceBaseList, setResourceBaseList] = useState<Resource[]>([]);
+  const toast = useToast();
+
+  useEffect(() => {
+    if (
+      process.env.NEXT_PUBLIC_AIRTABLE_KEY &&
```
narbs91, line 21: We change these guys removing `NEXT_PUBLIC` since i changed them up on vercel

gjsyme: That file is out anyway to do the caching. See brand new PR :grin:

gjsyme: Err. Commit. Whatever. If you've done your work already then I'm going to move it to a real PR.

src/lib/airtable.ts
gjsyme: believe this was a previous comment, and no, airtable sdk will create an empty object if there's nothing found (from what i found in their source that I think i pasted into that question)

src/utils/useAirtableResources.ts
narbs91: i've had my run through with it, if you're good to go we can turn it into a for realz PR hehe

gjsyme: ehhh it's failing on proptypes

narbs91:
So i took a look at the props code you were mentioning Discord @gjsyme and didn't see anything that stood out as "bad" in my eyes. The way you did it is the same way i would have done it wrt to the typing.

As an aside i was messing around with the Ui a bit and landed on something that looks like this:

Desktop

image

Mobile

image


Basically i slimmed it down to the bare essentials and used a nifty [url preview library](https://www.npmjs.com/package/@dhaiwat10/react-link-preview) our friend @Dhaiwat10 created 🔥 🔥 🔥 

I think it's cleaner and gives it a bit of lipstick to stand out though the direction for it is totally still up in the air

narbs91:
Plus later on we can also introduce a search bar uptop where we can search via the tags (where i've clumped the categories and tags together in the above screenshots)

narbs91:
I don't think i can push onto your branch but this is the code I adjusted in `ResourceCard/index.tsx` to get it to look like the above (rough code that still needs a bit of styling applied to look the very best):

(need to `yarn add @dhaiwat10/react-link-preview`)

```javascript
import { LinkPreview } from '@dhaiwat10/react-link-preview';
....
function KnowledgeCard(props: { data: Resource }) {

  const kbRecord = props.data.fields;

  return (
    <WrapItem maxW={{ base: '100%', md: '800px' }}>
      <VStack>
        <LinkPreview
          url={kbRecord.Source}
          width="500px"
          height="500px"
          descriptionLength={250}
        />
        <Author authors={kbRecord.extendedAuthors} />
        <HStack>
          {kbRecord.Category ? <Tag>{kbRecord.Category}</Tag> : null}
          {kbRecord.Tags?.map((tag: string, index: number) => (
            <Badge key={`tag-${index}`}>{tag}</Badge>
          ))}
          <LevelBadge levelName={kbRecord.Level} />
        </HStack>
      </VStack>
    </WrapItem>
  );
}
```

narbs91:
Oh looks like i can push, sweet hehe

gjsyme:
This is great. The visuals definitely needed some love and I think this fits the bill. I agree that filtering needs to be added down the road but I'd like to see this go live ASAP so I'm not worrying about it yet.

narbs91:
> This is great. The visuals definitely needed some love and I think this fits the bill. I agree that filtering needs to be added down the road but I'd like to see this go live ASAP so I'm not worrying about it yet. It might be nice to have a filler for those that don't have a visual in the cards.

Sweet, glad you like it :), yeah the filter is more of a nice to have atm but the overall page is good enough (minus some styling) for a MVP.  That's a good question around the default image, i don't know how to do that but maybe @Dhaiwat10 can give us the super secret details 👀

narbs91:
@gjsyme did you want me to help out with the styling and default image or are you going to take it over the finish line?

narbs91:
@gjsyme added the styling to the page and some stuff around a11y; however, didn't figure out the default image yet (not that i'm stuck just didn't have the time to dig into it)

Dhaiwat10:
Yoo you guys are using my library? 😂 Love that

Dhaiwat10:
I spotted some weird blank images too. It's been a long time since I looked into my library's source code. I'm gonna have a look at it & push a patch

Dhaiwat10:
Awesome work btw, love the look of this already!

kempsterrrr:
❤️  that this is being worked on already !! Can I suggest we don't show the un-curated view publicly just incase something makes it in there we really don't want going out on our website 😨 

Are y'all planning to include a way to filter/organise/group the resource to make it easier for users to find what they are looking for? I think that's important if we're going to push something out otherwise I 'd worry that it's not going to address the problem the resource base seeks to address - making it easier to find content you need.

narbs91:
> Yoo you guys are using my library? 😂 Love that

It was too good an opportunity to pass up using a D_D members library to spice up the website plus i really digged it anyways lol

narbs91:
> ❤️ that this is being worked on already !! Can I suggest we don't show the un-curated view publicly just incase something makes it in there we really don't want going out on our website 😨
> 
> Are y'all planning to include a way to filter/organise/group the resource to make it easier for users to find what they are looking for? I think that's important if we're going to push something out otherwise I 'd worry that it's not going to address the problem the resource base seeks to address - making it easier to find content you need.

I think it's pretty straight forward to default to the un-curated list but TBH i don't really know what curated means.  It could mean that it's curated to my liking which could signal that somehow the website has gotten a hold of my data and curating a list of articles (which isn't the case).  Basically we don't want to give the wrong impression to people (though i may be over exaggerating on my part :P)

With regards to filtering, I do believe @gjsyme and I may have spoken on the side about it but it's definitely on the radar.  On top of that, atm it would seem we're fetching the world from airtable and displaying it on the site; however, i don't think that's going to work in the long term with the amount of article churn we're expecting in the upcoming seasons.  Basically the page performance will take a hit when the load starts to increase and result in a negative user experience. We should think about introducing either pagination or a load more button with a cap of how many results per-fetch we do.

gjsyme:
We previously had the notion of draft or the like, which was meant for standing between a new submission and going live. That seems to no longer be the case. If our only means of filtering is the Curated boolean, then I agree, we should only have those which have been curated go on the page.
Filtering on the page, to me, makes the most sense for limiting number of views (in part because our API query is simple for the purposes of caching).

gjsyme:
I worked in the recommendations from @narbs91 (if you had put those changes on something, I couldn't find them) and called in @Dhaiwat10 's package / opened an issue for him to resolve the not-falsey but also not-functional responses that the proxy sometimes gets.
This does have the potential to jam us if the not-for-production proxy doesn't work out. Airtablle has some pictures worked in, but I'm not sure if this counts against our image calls. If so, would be something to proxy for a very long time (assuming vercell would hold that much data).

Dhaiwat10:
Hey. I just released a new version of the package and in the latest version you can pass in the fallback image src like this:

```
<LinkPreview url='someurl' fallbackImageSrc='src-to-your-image' />

**Comment by kempsterrrr** [comment]
> I think it's pretty straight forward to default to the un-curated list but TBH i don't really know what curated means

At the moment this is just a manual step in Airtable where we're explicitly accepting that something shows up in the public view. It's here to make sure we maintain a base layer of quality of the resources being included and also that nothing NSFW or in any other way harmful makes it into the the list. IMHO the curated view is the only view that should ever be public and the end user shouldn't even be need to be aware there is a difference between curated and not curated.

Regarding filtering/organising etc. this info on the page, maybe the benchmark should be what is provided out of the box from the Airtable views:

- List: https://airtable.com/shrzgqiMiHE18Iy9O/tbljejdzelezqT0W7
- Grid: https://airtable.com/shrM4PglRR3BAs3T9

I.e. people should be able to filter based on all the available data points and potentially free text search as well. That said, rather than asking them to do that we could alternatively have a structure to the page that groups content together.

Terrible implementation but something like:

- ethereum
  - beginner 
  - intermediate
  - advanced
- Solana
  - beginner 
  - intermediate
  - advanced

I appreciate that is a non-trivial task. Maybe whilst working on something more custom embedding the Airtable public view that gives those out of the box is a good interim solution.

See example below the fold on this website - https://blockchain-comparison.com/blockchain-protocols/

**Comment by gjsyme** [comment]
I feel like the default airtable view looks like the 90s wants its excel back. I'm down to continue working on making something web-worthy.
Can any of this be moved to a "fast follow" to borrow a term someone was throwing out on a voice call? Right now we have nothing in place.

**Comment by narbs91** [comment]
I think we can probably table the search/filter to a new issue as a fast follow and call the curated list as the MVP.  Filtering is always a bit of cluster to get get "right so it'll require some brainstorming on our end

**Comment by gjsyme** [comment]
Working through abusing the patience of @Dhaiwat10 as a very needy customer, but also making some iterative improvements as we go. @kempsterrrr the ability to view the non-curated data has been removed.

I'd be open to any "this is the way" type examples of how to filter from a list, but I'm just thinking generate a set of buttons/tags that you can click-to-toggle based upon the values associated with the curated resources.

**Comment by Neumand** [comment]
@gjsyme I stumbled across a pretty cool dev portfolio last night that implements the filtering based on the button/tags idea.

See: https://theodorusclarence.com/library

It’s open-source so feel free to check out the implementation.

**Comment by Dhaiwat10** [comment]
Just had a look at the Vercel deployment. Looking good!

**Comment by narbs91** [comment]
Looks good so far to me visually too, just some minor styling things wrt to padding on the titles and patching up mobile view a bit:

image


Haven't had a chance to look at the code yet but will try to make some time to have a look through today

File: src/components/ResourceCard/index.tsx [hunk 16]
```diff
@@ -0,0 +1,194 @@
...
+  const { t } = useTranslation();
+
+  if (!authors || authors.length === 0) return null;
+
+  return (
+    <VStack>
+      {authors.map((author) => (
+        <Flex key={author.name} align="center" justify="space-between">
+          {author.twitter ? (
+            <Link href={`https://twitter.com/${author.twitter}`}>
```
narbs91, line 60: I would say to add the `isExternal` prop here too to let the user know they'll have a new window open

src/components/ResourceCard/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+  return (
+    <Button variant="solid" leftIcon={buttonIcon}>
+      {buttonText}
+    </Button>
+  );
+};
+
+const FallBackImage = () => (
+  <VStack>
+    <Text>Image Unavailable</Text>
```
narbs91, line 122: [nit pick] these should be in the loc file

src/components/ResourceCard/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+function ResourceCard(props: { data: Resource }) {
+  const { t } = useTranslation();
+
+  const kbRecord = props.data.fields;
+  const kbDetails = renderResourceDetails(kbRecord);
+  const fallbackImage = FallBackImage();
+
+  return (
+    <WrapItem maxW={{ base: '100%', md: '300px' }}>
+      {/* <VStack minW={{base: '300px'}} maxW={{base: '100%', md: '300px'}}> */}
```
narbs91, line 178: [nit pick] commented out code

src/components/ResourceCard/index.tsx
```diff
@@ -0,0 +1,194 @@
...
+      <VStack w={{ base: '100%', sm: '30em', md: '300px' }}>
```
narbs91, line 179: Is there a reason we're mixing and matching units here?

gjsyme: i believe that the intent was to match the chakra measure of how large a sm screen was as 100% wasn't achieving the desired effect. but it's been a few days so there may be some data loss

src/types/airtable.ts
```diff
@@ -0,0 +1,51 @@
+// simplified airtable objects that work for our purposes
+export interface AirtableResponse {
+  _rawJson: string;
+}
+export interface Author {
+  id: string;
+  fields: {
+    Name: string;
+    'Developer DAO Member': boolean;
```
narbs91, line 9: I've never seen this style of syntax before, was there a special reason you went this route versus a regular variable declaration?

src/components/ResourceCard/index.tsx
narbs91: As well as a span with a visually hidden text that notifies a screen reader what this link represents and that it'll open in another window.  You can add this to the global.css file:

```javascript
.visuallyhidden {
  clip: rect(0, 0, 0, 0);
  clip-path: inset(50%);
  height: 1px;
  width: 1px;
  padding: 0;
  overflow: hidden;
  position: absolute !important;
  white-space: nowrap;
}
```

Then we can progressively add this type of thing to the other links we have as things go forward

src/lib/airtable.ts
```diff
@@ -0,0 +1,123 @@
...
+
+import {
+  Author,
+  Blockchain,
+  Category,
+  Resource,
+  Tag,
+} from './../types/airtable';
+
+// const airTable = new Airtable({ apiKey: process.env.NEXT_PUBLIC_AIRTABLE_READONLY_KEY });
```
narbs91, line 11: [nit pick] commented out code

src/components/ResourceCard/index.tsx
narbs91: ah i see, you can try using the vertical height and width units instead (i find they work better):  `vh` and `vw` respectively.  So `50vw` would mean half the windows width

src/types/airtable.ts
gjsyme: if you mean the content of 'fields' it's due to having to match the capitalized / spaced values coming off Airtable's API. otherwise, i might make an additional class to represent AirtableFields but didn't think it was necessry. i may very well be wrong.

gjsyme:
Last iteration: update dhaiwat dependency and resolve the common.json conflict.

Dhaiwat10:
Think we can merge this in. Looking good

Dhaiwat10:
We can, of course, have the design guild have a look at this later and get some design iterations going.

narbs91:
@with-heart @codingwithmanny Do we want to keep this PR?  This is actually something educational and aligns with the direction of the DAO.  Plus @gjsyme, @Dhaiwat10 and @Neumand have done a decent bit of work to get it to where it's at.

CC: @kempsterrrr

codingwithmanny:
@narbs91 kind of feel this could be put on hold, because some of it _MAY_ translate to the new website.

Dhaiwat10:
@narbs91 @codingwithmanny I think if we redesign this to be in-line with the new design scheme of the website, we can merge this in. In fact, we definitely want to have this on our website.

narbs91:
Yeah that's what my thinking around it was too. The content fits the bill of what the DAO wants to push out , great learning sources.  We can probs keep it open for the moment and readjust when the simplified landing page is pushed out.

PR #162, opened 2021-11-03
Feature: Testnet local environment

### LINKED ISSUE #155: Make local and commit deployments point to testnet for minting
### Context / Background / How This Idea Came To Fruition

After completion of #147 we realized that the minting functionality is always pointing to mainnet regardless of where the application is running.  This poses a hinderance wrt to testing locally and in a test environment when trying to actually mint since in the current form it will cost real ETH to preform. 

### Problem / Opportunity

With this change, developers working on future Ethereum blockchain features will connect to the Rinkeby testnet when working locally and testing on each per-commit deploy

### Proposed Solution / How It Can Be Improved

We will solve this issue by having an env based config to point to the Rinkeby testnet locally and on each pre-commit deploy while remaining mainnet in PROD(exact implementation is left to the developer).  The current Infura id can be used on all environments but we have to make code changes in [DirectMint](https://github.com/Developer-DAO/developerdao.com/blob/main/src/components/DirectMint/DirectMint.tsx) among other spots to account for the Rinkeby testnet.   Furthermore, since the Infura ID is stored as an env variable on Vercel the implementer of this issue will need to update the README file to note that any developer working on the codebase will need to have the infura id sent to them by a developerdao.com team member.

Please use `0xbf114c4cbb4e70e6098589a918c84292cb40602a` as the DEVS contract address when coding for the testnet and `https://rinkeby.etherscan.io/tx/` as the ETHERSCAN link prefix for the verification link.

### Testing the changes

Testing the changes will require potentially updating the e2e tests once #154 is complete in addition to manual smoke testing the existing minting functionality.

If you require Rinkeby testnet ETH please use the following [Chainlink faucet](https://faucets.chain.link/rinkeby?_ga=2.175183204.735297823.1635568853-1702889432.1630879734) to obtain some.

### Additional Information

Please ping the developerdao.com team in the Discord channel to ask for the infura ID before starting work on this issue

kavimaluskam:
### What does it do?

Use environment based constants to point home page and direct mint to test-net on local developement.

- Move all hardcoded network configuration to `src/utils/DeveloperDaoConstants.ts`, which can be switched from main-net to test-net based on environment.
- Test case looks good.
- Updating homepage at the same time to make the `test-net` environment more in-sync.
- Updating footer link at the same time to make the `test-net` environment more in-sync.
- Only confusion is now `https://developerdao.vercel.app/` is still pointing to production / main-net environment.

`process.env.NODE_ENV`, controls most network related constants in `src/utils/DeveloperDaoConstants.ts`.

### Relevant screenshots/gifs

Screenshot 2021-11-03 at 11 32 12 AM
Screenshot 2021-11-03 at 11 32 29 AM

### Does it close any issues?

Focusing on issue #155 

Closes #155


src/components/DirectMint/DirectMint.tsx
```diff
@@ -268,7 +270,12 @@ const DirectMint = ({ developerId }: DirectMintProps) => {
         </Modal>
       )}
 
-      {networkError && <Text color="red">{t('ethereumNetworkPrompt')}</Text>}
+      {networkError && NODE_ENV === 'development' && (
+        <Text color="red">{t('ethereumDevNetworkPrompt')}</Text>
+      )}
+      {networkError && NODE_ENV === 'production' && (
+        <Text color="red">{t('ethereumNetworkPrompt')}</Text>
+      )}
```
with-heart, line 278: I feel like it'd be more clear to combine these two related cases together:

```tsx
{
  networkError && (
    <Text color="red">
      {t(
        NODE_ENV === 'development'
          ? 'ethereumDevNetworkPrompt'
          : 'ethereumNetworkPrompt',
      )}
    </Text>
  )
}
```

src/utils/DeveloperDaoConstants.ts
```diff
@@ -1,13 +1,35 @@
...
 export const SITE_URL = 'https://developerdao.com';
 export const ERROR_CODE_TX_REJECTED_BY_USER = 4001;
-export const MAINNET_NETWORK_ID = 1;
+
 export const TOKEN_FINDER_URL = 'https://developerdao.vercel.app/';
-export const ETHERSCAN_TX_URL = 'https://etherscan.io/tx/';
+export const ETHERSCAN_TX_URL =
+  NODE_ENV === 'production'
+    ? 'https://etherscan.io/tx/'
+    : 'https://rinkeby.etherscan.io/tx/';
```
with-heart, line 32: This is not really related to your PR but seeing this makes me feel like we should start using the [Next.js Environment Variables feature](https://nextjs.org/docs/basic-features/environment-variables) because most of these aren't constants.

with-heart (approved):

with-heart:
> Only confusion is now https://developerdao.vercel.app/ is still pointing to production / main-net environment.

I'm not sure we have a good way around this with Vercel. Preview deploys are meant to mimic prod, not to be a test env.

src/utils/DeveloperDaoConstants.ts
kavimaluskam: Yeah this would even be more clear.

kavimaluskam:
> > Only confusion is now https://developerdao.vercel.app/ is still pointing to production / main-net environment.
> 
> I'm not sure we have a good way around this with Vercel. Preview deploys are meant to mimic prod, not to be a test env.

I think it's good to have Vercel preview to be mimic prod, just the current https://developerdao.vercel.app/ is showing another version / website 👀 
Screenshot 2021-11-03 at 2 54 41 PM

with-heart:
OH I misread that URL! I was thinking it was our site's vercel URL, not the separate app.

kavimaluskam:
> OH I misread that URL! I was thinking it was our site's vercel URL, not the separate app.

I'd have provide more context for this.

The https://developerdao.vercel.app/ is referring to the TOKEN_FINDER_URL in direct mint page. Which looks to be pointing to another application by @tfmoliveira, providing list of available IDs. 

Maybe we shall discuss if we'd provide a similar functionality and replace the standalone application? 👀

src/components/DirectMint/DirectMint.tsx
```diff
@@ -65,7 +67,7 @@ const DirectMint = ({ developerId }: DirectMintProps) => {
 
   useEffect(() => {
     const web3Modal = new Web3Modal({
-      network: 'mainnet',
+      network: DEVELOPER_DAO_CONTRACT_NETWORK,
       cacheProvider: false,
       providerOptions,
```
narbs91, line 72: So overall things look great testing with MetaMask; however, in the case of walletConnect that won't work with the current setup if there's no NEXT_PUBLIC_INFURA_ID set up on the developers machine.  So let's do a couple of things before we wrap up this PR:

1.  Let's add a sample.env file to the root of the project and have it look just like this
```javascript
NEXT_PUBLIC_INFURA_ID=''
```
2.  Add a blurb in the `Developing` section of the readme that people who are testing/building blockchain related features to connect with MetaMask as a first option (as that doesn't require a infura_id).  If they want to test via walletConnect then they have to create an infura account themselves and add their personal infura_id to a .env file they create based on the sample we provide.  I'm also ok with not mentioning the walletConnect part at all and steer blockchain interaction testing towards MetaMask (though that might detract from adding more ways to connect through various providers in the future).

narbs91 (changes requested):

narbs91:
Also looks like we're having a failure with the directMinting e2e test fyi

src/utils/DeveloperDaoConstants.ts
codingwithmanny: Yeah I'm leaning towards more and more for the ENV vars because for the GitHub Actions, we're going to need to change almost everything to fit that scenario.

Thoughts on maybe a placeholder in the meantime:

```js
const DEVELOPER_DAO_CONTRACT = process.env.DEVELOPER_DAO_CONTRACT || (NODE_ENV === 'production'
    ? '0x25ed58c027921E14D86380eA2646E3a1B5C55A8b'
    : '0xbf114c4cbb4e70e6098589a918c84292cb40602a');
```

kavimaluskam: Actually I prefer ENV VARs for as well. I think env variables is now very common for secrets and setting configuration for application with docker, cloud instance, and CI. 

Happy to do the change to use env var, but probably need to involve someone with vercel access.

src/components/DirectMint/DirectMint.tsx
kavimaluskam: Aw yeah I've forgot the testing about walletConnect. 

Thinking shall we complete the update to use ENV in this PR?

src/utils/DeveloperDaoConstants.ts
with-heart: > While certainly possible to do I would argue that the typical usage for env variables that i've seen is usually towards storing secrets; however, even that is probably better served using a secrets manager TBH.

That's funny! My experience has been sort of the opposite. Env vars stored in checked-in files like `.env.development` are usually for differing URLs and such, with secrets being stored as env vars in something like Vercel's env var secrets manager.

src/components/DirectMint/DirectMint.tsx
kavimaluskam: And kinda curious how NEXT_PUBLIC_INFURA_ID is currently setup in the production enviornment? 🤔

src/utils/DeveloperDaoConstants.ts
codingwithmanny: @with-heart samesies

> > While certainly possible to do I would argue that the typical usage for env variables that i've seen is usually towards storing secrets; however, even that is probably better served using a secrets manager TBH.
> 
> That's funny! My experience has been sort of the opposite. Env vars stored in checked-in files like `.env.development` are usually for differing URLs and such, with secrets being stored as env vars in something like Vercel's env var secrets manager.

src/components/DirectMint/DirectMint.tsx
codingwithmanny: Agree with @narbs91 

@kavimaluskam like as in the admin dashboard or how it exposes the env vars?

> And kinda curious how NEXT_PUBLIC_INFURA_ID is currently setup in the production enviornment? 🤔

kavimaluskam: @codingwithmanny so there is one production `NEXT_PUBLIC_INFURA_ID` in vercel ?

And i am not very familiar with INFURA_ID, is it like some SDK ID? Can we have one for development ?

codingwithmanny: @kavimaluskam Technically anyone can sign up for one, and I think we're using someone's 🤷 .

I have to go through the process again, but the site is https://infura.io.

I believe it would work as the same process as signing up for an etherscan API key, it just allows you to do specific types of calls to contracts that would normally be done in your browser with metamask (even then there's a rate limit).

src/utils/DeveloperDaoConstants.ts
kavimaluskam: It sounds good for me as well! Just generally means ENV is a good shit.

kavimaluskam:
Just added an env. var. approach.

src/components/DirectMint/DirectMint.tsx
narbs91: Sorry i got to this a little late but just for clarity @kavimaluskam there is a `NEXT_PUBLIC_INFURA_ID` env variable set up in vercel with a valid (at least for now hehe) infura_id.  Since that env variable isn't available for local builds of the app we want to have people set their own variable (i.e. in a .env file) and populate it with their own infura_id like Manny explained above.  Infura itself is a cloud provider for Ethereum nodes (and i think IPFS too but don't remember) so that people can easily connect to the blockchain without having to setup a node themselves

kavimaluskam: Cool! @narbs91 
Have added a line in the README to ask for using `.env.development` for the `NEXT_PUBLIC_INFURA_ID`.

The `NEXT_PUBLIC_INFURA_ID` is missing in `.env.production`, thinking if it can be overrided by Vercel ENV?

src/components/DirectMint/DirectMint.tsx
```diff
@@ -91,7 +93,7 @@ const DirectMint = ({ developerId }: DirectMintProps) => {
   };
 
   const _checkNetwork = (chainId: number) => {
-    if (chainId === MAINNET_NETWORK_ID) {
+    if (chainId === NETWORK_ID) {
```
narbs91, line 96: So we actually have to change this condition slightly since `NETWORK_ID` is actually a string and not a number:
```javascript
 if (chainId.toString() === NETWORK_ID) {
      return true;
    }
```

src/utils/DeveloperDaoConstants.ts
```diff
@@ -1,13 +1,31 @@
 export const DEVELOPER_DAO_CONTRACT =
-  '0x25ed58c027921E14D86380eA2646E3a1B5C55A8b';
+  process.env.NEXT_PUBLIC_DEVELOPER_DAO_CONTRACT ||
```
narbs91, line 2: You actually don't need the extra conditionals here since you have specified a `env.development` which will be the profile that'll be loaded when someone runs `yarn dev`.  I would say it should be fine to remove these fallbacks

.env.development
```diff
@@ -0,0 +1,9 @@
+NEXT_PUBLIC_DEVELOPER_DAO_CONTRACT=0xbf114c4cbb4e70e6098589a918c84292cb40602a
```
narbs91, line 1: So i ended up adding these key/values in vercel in two flavours:  one set for development/preview and another for production.  I re-ran the deploy job and it ended up picking up the env.development variables on the preview deploy (yay!) which solves our test vs prod issue.  So what that'll mean is that we can actually test minting on the preview builds against rinkeby and leave production as is with the legit minting

.env.development
```diff
@@ -0,0 +1,9 @@
...
+NEXT_PUBLIC_DEVELOPER_DAO_CONTRACT_NETWORK=rinkeby
+NEXT_PUBLIC_DEVELOPER_DAO_CONTRACT_NETWORK_PROVIDER=rinkeby
+NEXT_PUBLIC_NETWORK_ID=4
+NEXT_PUBLIC_OPENSEA_DIRECT_LINK_PREFIX=https://testnets.opensea.io/assets/$DEVELOPER_DAO_CONTRACT
```
narbs91, line 5: Can you switch `$DEVELOPER_DAO_CONTRACT` at the end here with the actual contract as I don't know if vercel can interpolate env variables like that

.env.production
```diff
@@ -0,0 +1,9 @@
+NEXT_PUBLIC_DEVELOPER_DAO_CONTRACT=0x25ed58c027921E14D86380eA2646E3a1B5C55A8b
+NEXT_PUBLIC_DEVELOPER_DAO_CONTRACT_NETWORK=mainnet
+NEXT_PUBLIC_DEVELOPER_DAO_CONTRACT_NETWORK_PROVIDER=homestead
+NEXT_PUBLIC_NETWORK_ID=1
+NEXT_PUBLIC_OPENSEA_DIRECT_LINK_PREFIX=https://testnets.opensea.io/assets/$DEVELOPER_DAO_CONTRACT
```
narbs91, line 5: Same comment as above here

narbs91 (changes requested):
Looks like we're really close to closing this out after this last round of feedback @kavimaluskam ! Great stuff 👌

.env.development
kavimaluskam: I think it's handled by Next.js? 
https://nextjs.org/docs/basic-features/environment-variables

But yeah, maybe it's better to make a full string in vercel and make both .env and vercel consistent.

src/utils/DeveloperDaoConstants.ts
```diff
@@ -1,13 +1,25 @@
 export const DEVELOPER_DAO_CONTRACT =
-  '0x25ed58c027921E14D86380eA2646E3a1B5C55A8b';
+  process.env.NEXT_PUBLIC_DEVELOPER_DAO_CONTRACT;
```
narbs91, line 2: So the build seems to be failing complaining about this potentially not being a string, let's add `as string` to the end of these to make typescript happy

src/components/DirectMint/DirectMint.tsx
narbs91: Reminder that this still has to be done to get things to connect properly

.env.production
```diff
@@ -0,0 +1,9 @@
+NEXT_PUBLIC_DEVELOPER_DAO_CONTRACT=0x25ed58c027921E14D86380eA2646E3a1B5C55A8b
+NEXT_PUBLIC_DEVELOPER_DAO_CONTRACT_NETWORK=mainnet
+NEXT_PUBLIC_DEVELOPER_DAO_CONTRACT_NETWORK_PROVIDER=homestead
+NEXT_PUBLIC_NETWORK_ID=1
+NEXT_PUBLIC_OPENSEA_DIRECT_LINK_PREFIX=https://testnets.opensea.io/assets/0x25ed58c027921E14D86380eA2646E3a1B5C55A8b
+NEXT_PUBLIC_OPENSEA_COLLECTION_LINK=https://testnets.opensea.io/collection/devs-for-revolution-kexfxbcvp8
+NEXT_PUBLIC_ETHER_SCAN_LINK_PREFIX=https://rinkeby.etherscan.io/address
+NEXT_PUBLIC_ETHERSCAN_TX_URL=https://rinkeby.etherscan.io/tx/
```
narbs91, line 8: These guys are still point to their respective test urls, let's change these to point to their prod counter parts

narbs91 (changes requested):

src/components/DirectMint/DirectMint.tsx
kavimaluskam: Oh I've missed it, fixed.

README.md
```diff
@@ -84,6 +84,7 @@ nvm install; # to install the version in .nvmrc
 2. Install dependencies by running `yarn`
 3. Start the development server with `yarn dev`
 4. Open development site by going to `http:localhost:3000`
+5. To test Walletconnect functionality, copy `.env.development` to `.env.local` and fill in your `INFURA_ID`.
```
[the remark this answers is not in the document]
kavimaluskam: sure updated.

narbs91:
Great stuff @kavimaluskam !  So looks like we have fully functional Rinkeby minting on our preview/per-commit builds 🥳 .  I've added the remaining production environment variables to Vercel so we should be good to go after you resolve your conflicts with the master branch

kavimaluskam:
@narbs91 looks the new `useDevNFTSupply` is not working in testnet 🤔 let me look into this.

kavimaluskam:
@narbs91 Solved and updated the `useDevNFTSupply` with testnet functionality 👍

narbs91 (approved):
YAY!  Awesome work @kavimaluskam this is really going to help us out in the long term when we're developing blockchain oriented features with a legit "stagging"/"int" environment

kavimaluskam:
Thanks for your help on this PR mates 🎉🎉

PR #168, opened 2021-11-04
chore(missing-translations): make locale single select

with-heart:
This PR changes the "locale" field of the "missing translations" issue form to be single-select rather than multi-select.


narbs91 (approved):
LGTM!

PR #181, opened 2021-11-14
Fix couple misused characters in zh-TW

amazingandyyy:
### What does it do?

Fix typos and some misused characters

- The first version of translation has many simplified Chinese characters, this is due to many online translators/translating engine nowadays having difficulty to use decent "transitional Chinese"
- I am native Taiwanese who can correct them

ignore `.idea` file, IntelliJ is a popular editors and we should ignore `.idea`(the IntelliJ confirmation file) if possible

### Does it close any issues?

No issue

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,3 +23,6 @@ yarn-debug.log*
 yarn-error.log*
 
 .next
+
+# editors
+.idea
--- a/public/locales/zh-TW/common.json
+++ b/public/locales/zh-TW/common.json
@@ -21,7 +21,7 @@
   "enterDeveloperId": "請輸入$t(title)編號",
   "hosting": "托管於:",
   "copyLinkToNFT": "複製 NFT 鏈接",
-  "linkCopied": "鏈接已複製到剪貼板",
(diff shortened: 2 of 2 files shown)


Dhaiwat10:
Thanks @amazingandyyy!

kavimaluskam:
Thanks @amazingandyyy would you mind to update the ZH-HK file as well? 

I think it make sense to in sync both traditional files.

with-heart:
I think that makes sense @kavimaluskam! Would be great to tackle both.

amazingandyyy:
Altho they are both traditional Chinese, but in reality, HK use different terms and characters than TW's

As a Taiwanese, I suggest having members from HK to translate for zh-HK for respect HK's community

amazingandyyy:
I think we can merge this first and for put zh-HK correction on another PR if anyone is capable with it.

with-heart:
@amazingandyyy That's fair! We'll let an HK'er translate `zh-HK`.

@allcontributors please add @amazingandyyy for translation

PR #182, opened 2021-11-14
chore: run prettier after generating types

with-heart:
This PR modifies the `generate-types` script to run `prettier` on the generated type files in `types/ethers-contracts`. 

This change resolves an issue where running `yarn` caused types to be generated that weren't formatted with `prettier`, so we ended up with unstaged changes to those files where the changes were purely the missing formatting.


Dhaiwat10 (approved):
Nice catch

PR #185, opened 2021-11-15
Added conditional render for OpenSea Link - removed for now

### LINKED ISSUE #184: Remove link to Opensea
### Context / Background / How This Idea Came To Fruition

From Nader's [message](https://discord.com/channels/883478451850473483/884852584941232169/909666569846669363)

image

### Problem / Opportunity

> we are not trying to encourage people to buy the tokens on the secondary, at least until we launch the new Governance token.

### Proposed Solution / How It Can Be Improved

Remove the link to Opensea from the footer

### Testing the changes

-

### Additional Information

_No response_

namanmanchanda09:
### What does it do?
Added a variable for conditional render of OpenSea link in footer. Updated the tests in `_tests_/Footer/index.test.jsx` accordingly.

A simple conditional render according to the variable **openseaConfigFlag** was added.

### Does it close any issues?
Closes #184

### DIFF
--- a/_tests_/Footer/index.test.jsx
+++ b/_tests_/Footer/index.test.jsx
@@ -6,13 +6,11 @@ describe('Footer', () => {
   it('Renders the icons', () => {
     render(<Footer />);
 
-    const openSea = screen.getByTitle('OpenSea');
     const etherscan = screen.getByTitle('Etherscan');
     const twitter = screen.getByTitle('Twitter');
     const discord = screen.getByTitle('Discord');
     const github = screen.getByTitle('GitHub');
 
-    expect(openSea).toBeInTheDocument();
     expect(etherscan).toBeInTheDocument();
     expect(twitter).toBeInTheDocument();
     expect(discord).toBeInTheDocument();
--- a/src/components/Footer/index.tsx
+++ b/src/components/Footer/index.tsx
@@ -17,6 +17,7 @@ import {
 
 function Footer() {
   const { t } = useTranslation();
+  const openseaConfigFlag = false;
 
   return (
     <chakra.footer borderTop="1px solid" borderColor="gray.200" py={10}>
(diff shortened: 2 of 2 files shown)


Dhaiwat10 (approved):
Thanks, LGTM!

I have one suggestion but it might be overkill for this case. Here it goes anyways: maybe we could have the flag as an `.env` entry rather than hardcoded inside the code. With the current setup, if we want to bring the link back we will also need to manually update the test file. Instead of that, we could have the flag loaded from the `.env` and read it inside the test file too.

namanmanchanda09:
All right. Want me to do it?

Dhaiwat10:
> All right. Want me to do it?

Let's wait for one or more reviews and see what they think

with-heart:
imo we should just remove the link completely (including the `NEXT_PUBLIC_OPENSEA_*` vars in the `.env` files and `OPENSEA_*` vars in `DeveloperDaoConstants.ts`). It'd be trivial to later revert this PR to add it all back.

Conditionally rendering behind a flag is just a slightly fancier equivalent of commenting out the code 😀

kavimaluskam:
> imo we should just remove the link completely (including the `NEXT_PUBLIC_OPENSEA_*` vars in the `.env` files and `OPENSEA_*` vars in `DeveloperDaoConstants.ts`). It'd be trivial to later revert this PR to add it all back.
> 
> Conditionally rendering behind a flag is just a slightly fancier equivalent of commenting out the code 😀

Sounds a better approach as it make good use of the `.env`, and 1 less external dependency on the flag.

Just have to test and make sure the empty `<li> </li>` doesn't affect the UI, or can move the dependency to the `OPENSEA_` constant itself.

namanmanchanda09:
> > imo we should just remove the link completely (including the `NEXT_PUBLIC_OPENSEA_*` vars in the `.env` files and `OPENSEA_*` vars in `DeveloperDaoConstants.ts`). It'd be trivial to later revert this PR to add it all back.
> > Conditionally rendering behind a flag is just a slightly fancier equivalent of commenting out the code 😀
> 
> Sounds a better approach as it make good use of the `.env`, and 1 less external dependency on the flag.
> 
> Just have to test and make sure the empty `<li> </li>` doesn't affect the UI, or can move the dependency to the `OPENSEA_` constant itself.

namanmanchanda09:
Sorry - closed by mistake. Reopened now. The above comments make sense. @narbs91 suggested to add a config-flag for the OPENSEA link. Lmk if to make the above suggested changes.

narbs91:
The config-flag wasn't a must and looking at the changes now it does look pretty trivial to just rip out and put back later (that's what GitHub history is for lol).  One other thing is that there's a button just below the NFT on the landing page that directs to the NFT itself on opensea.  Were you also thinking of having that removed too @dabit3 ?

PR #197, opened 2021-11-20
Adds localPath fix to address missing locale issues on build

narbs91:
### What does it do?

Fixes issue with potential build issues on Vercel having to do with localePaths being missing

This was a long standing issue that only got recently resolved in a previous issue that was closed due to a change in direction of the website.  This is a necessary change for future localized paths that may end up on the site

N/A

### Relevant screenshots/gifs

N/A

### Does it close any issues?
N/A

Closes #...

### DIFF
--- a/next-i18next.config.js
+++ b/next-i18next.config.js
@@ -1,3 +1,5 @@
+const path = require('path');
+
 module.exports = {
   i18n: {
     defaultLocale: 'en',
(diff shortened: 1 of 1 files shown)


codingwithmanny:
@narbs91 we still doing this or this going to be archived as well?

narbs91:
> @narbs91 we still doing this or this going to be archived as well?

This i saw as just general maintenance to ensure localization works properly going forward (assuming we use the same localization library)

Dhaiwat10 (approved):
Good catch

PR #221, opened 2022-04-12
Update membership info based off the notion details

etr2460:
### What does it do?
As titled

In https://developerdao.notion.site/Getting-Started-with-Developer-DAO-2bddd332c51a4957b0b83f60f9fa4ebe we see that ddao is open for new members now. Let's update our website accordingly. Ideally this might use inline links to opensea, but i'm not 100% sure how that might play with the translations. So for now let's just update the content to match what's in the notion

to: @with-heart @Dhaiwat10

### DIFF
--- a/public/locales/en/common.json
+++ b/public/locales/en/common.json
@@ -35,7 +35,7 @@
   },
   "membership": {
     "title": "How do I join?",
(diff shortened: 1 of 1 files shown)


Dhaiwat10:
Thanks @etr2460 :) Unfortunately I dont have Vercel access, we'll have to wait on with-heart

PR #250, opened 2022-07-24
SEO component

### LINKED ISSUE #247: Create Reusable SEO Component
Currently handling all SEO globally in the app component.

task is to create a reusable SEO component as we expand beyond a single page.

ephcrat:
### What does it do?

Reusable SEO component 

Removed it from the app component on the pages folder and created a new file in a separate folder. Also, removed the now unused dependencies from the app component. 

It now uses the props "title", "description" and "url" with the values used on the app component set as default. I'm using i18 for the first time not sure if this is the correct way to pass the props. Please let me know if there's something wrong or if I'm missing something. All feedback is appreciated.

### Does it close any issues?

Closes #247


mhmulford0:
thanks for this! We just released the version of the site. Looks like this caused some merge conflicts. Please have a look when you can, I would love to get your work added.

ephcrat:
> thanks for this! We just released the version of the site. Looks like this caused some merge conflicts. Please have a look when you can, I would love to get your work added.
> 
> Also please pick up the task here: https://app.dework.xyz/developer-dao/developerdaocom

I'm unable to pick up the task :( doesn't appear the "I'm interested" option for that task.

mhmulford0:
Are you a member of the DAO? It looks like it should be open to anyone with the default dev role in the discord. @ephcrat

